PR #6, opened 2016-08-29
Max/message box component

maxwellsmart84:
Here is my first pull request for RhinoStyles:

Changes Made:

+ Added new library, Jack Moore's widely used 'autosize' (npm install autosize), this has been    added to our package.json as a dependency.

+ I based the component off a js fiddle found here http://jsfiddle.net/FrancescoCioria/hL842e3d/

+ Added component to index, added examples and playground with credit to the autosize author.

+ Added an inline style (line 79 in MessageBox.jsx) that sets the white-space css property to normal, this is to override the white-space property being applied to the placeholder attribute which was breaking the line-break functionality of the autosizer (fix from last PR).


mcraiganthony:
1) Add a **maxHeight** prop. Default value would be 'none'. If number is passed, add max-height style to **messageBoxStyle**. 

2) Change **displayName**

3) Verify that it works in other browsers. You can use my iMac to check FF, Chrome, and IE11 on Windows. Please test using your iPhone as well - or other mobile device.

@igreulich @benbruning should the **Textarea** component have children propType like **MessageBox**? Seems like it should.

maxwellsmart84:
Per conversation with Craig - also adding children Proptypes to the Textarea component.

maxwellsmart84:
@mcraiganthony I fixed the firefox issue (red border) by removing the loading of all the props, its not needed I believe as the component works fine without it.

maxwellsmart84:
I forgot to mention that Craig put a fix in for the text area not indenting text when it reached the width of the textarea box.

mcraiganthony:
I'll take a look in a bit!

maxwellsmart84:
@mcraiganthony I know believe this is ready for merge - test on all browsers and made sure props worked as well.

igreulich:
@maxwellsmart84 - I love the new commit messages.

I would love to see smaller commits, though.

The idea behind this that we can revert the smaller piece, and not 4 changes, should something terrible happen. (that is not the case here, I am thinking more about production, etc.)

PR #15, opened 2017-01-12
Blake/build

BlakeGuilloud:
I've been deving in rhinostyle a bit lately- just minor tweaks to Selects / Dropdowns etc.. and I've found it's kind of a pain to work around the current build task + the eslint rules.. This doesn't by any means solve all of the eslint stuff, but it at least makes it less like the red wedding.. 

We can chat about this tomorrow, as I know specifically the react/no-unescaped-entities rule will be a controversy, but literally on every single xxApp.jsx file it is throwing an eslint error. Investigation makes me really wonder if this is a bug with the actual eslint plugin, as it is mis-recognizing things like `<code>type="primary"</code>` as HTML elements as opposed to just rendering the content.. If you think of a better way of handling this issue, let me know and I'll throw it in this PR..

Also- the build task was driving me crazy so I started listening to files in `components/*`.. Let me know if this is a problem. I haven't seen any problems with it, and it makes the dev environment a lot more consistent.


benbruning:
I think this is great.. I have no issues with it.  @igreulich @mcraiganthony ?

.eslintrc
```diff
@@ -52,13 +52,15 @@
     "arrow-body-style": 1,
     "import/prefer-default-export": 1,
     "import/extensions": 0,
+    "import/no-extraneous-dependencies": 0,
```
igreulich: We use `"import/no-extraneous-dependencies": ["error", {"devDependencies": true}],` in rhinofront, I think this should match.

igreulich:
@BlakeGuilloud Do you have `eslint` installed globally? If so, what version do you have installed?

igreulich:
It looks good to me. I will merge it in.

PR #18, opened 2017-01-27
Blake/input

BlakeGuilloud:


tken85:
this is related to https://github.com/rhinogram/rhinofront/pull/362

mcraiganthony:
Needs some work. Probably best to include these attributes in the playground example so that we can test on mobile.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

mcraiganthony:
LGTM

PR #19, opened 2017-01-31
Trimming inputs / textareas and message box

BlakeGuilloud:


src/scripts/components/Input.jsx
```diff
@@ -75,7 +75,7 @@ class Input extends React.Component {
     this.setState({ value: event.target.value });
 
     if (this.props.onChange && typeof (this.props.onChange === 'function')) {
-      this.props.onChange(event.target.id, event.target.value);
+      this.props.onChange(event.target.id, event.target.value.trim());
```
benbruning: Do we need to add the .trim to this.setState({ value: this.props.initialValue }); on the willmount's?

BlakeGuilloud: i'll investigate. good catch

BlakeGuilloud:
@mcraiganthony I want to make sure you are ok with this as well, as it has the UI implication of clearing out any whitespace that prefaces the first block of text in an input.

PR #21, opened 2017-02-07
updates to input message box and text area trimming methods

BlakeGuilloud:
### What should this PR do?
Prevent input components from jumping an empty whitespace while deleting input values.
This is caused by the way we are handling the state of the initialValue property in rhinofront. 

### What are the acceptance criteria?
1. Run an `npm link rhinostyle` of any instance on rhinofront and confirm that any Input, Textarea, and MessageBox components do not jump after typing two words, and deleting the second.

### Note
The jumping cursor on type="email" Input components is not a result of this feature.


BlakeGuilloud:
This should be go. I am going to handle the Avatar issue on rhinofront.

zslabs:
@BlakeGuilloud If the email input is still wonky, I think this will be fine to merge as-is. Let me know what you think.

BlakeGuilloud:
@zslabs I spent some time yesterday researching / debugging the email issues and could not find a solid solution.. Basically type="email" wants it's value to be oneWord.. I'm not sure what the solution to this would be.

Should be good to go once conflicts are resolved.

zslabs:
👍

PR #33, opened 2017-02-20
Zach/gsap dropdowns

zslabs:
- [x] I've tested in all browsers (IE11, Firefox, Safari, Chrome, and iOS)

### What should this PR do?
Switches out the CSS transition-based Dropdowns with GSAP. This impacts both the Dropdown and DropdownMultiselect components. As with the Modal and Cover work, we now have the following props we can listen to within the `<DropdownWrapper>`:

```js
onComplete: React.PropTypes.func,
onReverseComplete: React.PropTypes.func,
onReverseStart: React.PropTypes.func,
onStart: React.PropTypes.func,
```

This addition also changes how the DropdownMultiselect responds to clicks of a selectable item. Currently, when you select an item, the dropdown closes, which means you have to re-focus on the input to select another -- which is a bit frustrating since the purpose of the component is to multi-select. Now, the component responds to the `onClickOutside` action. Also, unlike normal dropdowns, focusing on the input used for filtering (if the dropdown is open) will not trigger a close action since we'd like users to be able to both interact with what's available in the list currently, filter down that list, and then select more without having to re-trigger the dropdown itself.

1. Attempt to interact with prop updates added
1. Make sure there are no style regressions.
1. Ensure the above changes to DropdownMultiselect works as intended.


zslabs:
@mcraiganthony Let me know if the style updates look alright to you.

PR #48, opened 2017-03-22
Zach/message style updates

zslabs:
### What should this PR do?
Swaps out the current messaging style for a more familiar "bubbly" effect that iOS has. I worked with @jpdrake on this one and have sent over the preview attached in the app to him as well. **I still can't test in IE through my Virtual machines for the app, but did verify the visuals in Rhinostyle.**

message-update


mcraiganthony:
My 2 cents ... the tail is preferable to the bubbles (they point to void space) and the left/right sides need a tad more padding (subjective - I could live with it as is).

zslabs:
@mcraiganthony I amended the PR to use the standard tails. Lemme know what you think.

zslabs:
screen shot 2017-03-24 at 1 06 55 pm

PR #49, opened 2017-03-23
Zach - Build updates + inline SVG

zslabs:
### What should this PR do?
While working to inline the SVG sprite as we did in [Rhinofront](https://github.com/rhinogram/rhinofront/pull/678) we can't approach it in the same way as we are using Metalsmith to generate these pages; and not Webpack + React. While adding in the sprite to be "includeable" by Metalsmith, I ran into an issue with sprite not being found during the `watch` command. This was due to how Nunjucks was being configured. I also saw this as an opportunity to bring in some of the logic from my framework to help us not need `metalsmith-changed` anymore and clean up the gulp tasks to be able to selectively start up a server without having to run through all of the build if you'd like. Happy to run through this.

1. Ensure pages are built as they are currently.
1. Make sure SVG is inlined at the top of the page onload without a separate request.
1. Make sure changes to files refresh properly.


mcraiganthony:
LGTM

PR #50, opened 2017-03-27
Zach - Panel animations

zslabs:
- [x] I've tested in all browsers (IE11, Firefox, Safari, Chrome, and iOS)

### What should this PR do?
Removes the CSS-based panel scaffolding logic for a GSAP-powered version. This only impacts the documentation site and is used as a precursor to work to be done within the app. 

Since RS is mobile-first, having the slide-out navigation show its "desktop-version" presented a visual regression since the styles would be parsed first and then the Javascript, creating a jittery loading experience. The way I combatted this was to show the page on desktop once the scaffolding had been figured out in GSAP. Since we are using `progress()` to immediately render the end-portion of the timeline, the transition is instantaneous to the user on subsequent page-loads and is only in-effect when the script is not cached yet. Since RS/RF are JavaScript based that need to be parsed to show the content, this felt like an adequate workaround for the benefits that GSAP will provide us when interacting with animation states within the app.

1. Ensure when you resize the browser the UI reacts as it does currently.

The one item I came across was that Edge is treating the load-state a bit differently than I would expect, but I'm always noticing a bit of weirdness in the current UI as well. Everything eventually loads fine, but there is a slight jitter that I did not notice in other browsers. Can show you whenever you have a sec. 

**Note:** I removed the opacity transition I showed earlier on-click and on-load to shorten the time the user is waiting for pages to be "interactable". The amount of time it takes between the request to the load the page and when it is shown was about `6.21 milliseconds`.


zslabs:
Closed. Using https://github.com/rhinogram/rhinostyle/pull/51 instead.

PR #57, opened 2017-04-12
Zach - dep updates

zslabs:
### What should this PR do?
* Updates all `package.json` dependencies (except for `metalsmith-in-place because that update is stupid
* Cleaned up a lot of eslint errors and enabled `eslint-loader` within Webpack to run on both the docs and dist build of RS
* Removed `react-gsap-enhancer` for native GSAP approach for toast notifications

1. Ensure toast notifications work as expected.
1. Make sure running `gulp dist:scripts` and `gulp docs:scripts` do not generate any eslint errors/warnings.

**This PR also includes updates to break out `React.PropTypes` into what `15.5.*` recommends. Because not all of our 3rd party components are ready for prime-time (and if they are not there's a crap-load of console warnings) I updated our code, but not the React versions so we can at least be proactive for when everything else is ready.**


zslabs:
@benbruning I've updated this PR to revert the `react` and `react-dom` versions as we chatted about. Let me know if all looks good on your end. Once this is in and a new version is released, I'll finish up the RF PR as well.

PR #66, opened 2017-05-10
Zach/forms

zslabs:

### What should this PR do?
* Removed inline styles from form, checkbox, and radio. Now using `<UtilityInlineGrid>` wherever possible.
* Updated `<RhinoSwitch>` component to be more mobile (touch) friendly (let me know if you agree with the color swap as well) and simplified code presentation.
* Adjusted list utilities. Also created `<UtilityList>` and `<UtilityListItem>` for unordered lists
* Favoring native `[disabled]` attribute for styling over `.is-disabled` wherever possible to reduce redundancy. Also updated classname to match other "state" classes.
* Cleaned up various form styles that were either not in use or were not using global vars
* Updated documentation for new form options
* Resolved scale issue on Windows IE11, Edge, and Firefox


zslabs:
@mcraiganthony This is ready for a formal review. Let me know when you have some time to run through this.

PR #71, opened 2017-06-20
Loading state for buttons

zslabs:

### What should this PR do?
Adds in a `loading` prop to the `<Button />` component that allows us to re-use the `<LoaderCircle />` component to show a loading-state.

buttons-loading

This PR also changes disabled states for primary, secondary, and danger to mimic the default and adds in a `xsmall` size to `<LoaderCircle />` for smaller buttons. This size was already in-use in the app in the typing indicator, so when merged that extra code can be deleted.

1. Adding `loading` prop should disable the button and show loading animation
1. `xsmall` size modifier in `<LoaderCircle />` component should show xsmall size.


mcraiganthony:
I prefer the expanding method as seen in the examples below - text stays, button wides, loader appears. However, I do realize that this affects layout and is a tradeoff. Thoughts @zslabs @jpdrake ?

http://lab.hakim.se/ladda/
https://codepen.io/jmalatia/pen/HkmaA

zslabs:
Yeah, I do love that effect, but I definitely agree with what you said about impacting layout; especially on smaller sizes.

src/scripts/components/Button.jsx
```diff
@@ -61,20 +76,23 @@ class Button extends React.Component {
...
     });
 
     let markup = '';
 
     if (route) {
       markup = (
-        <Link to={route} className={classes} onClick={this.handleClick} {...opts} title={this.props.title}>
+        <Link to={route} className={classes} onClick={this.handleClick} disabled={disabled || loading} {...opts} title={this.props.title}>
           <span className="button__text-wrapper">{this.props.children}</span>
+          {this.loadingRender()}
```
BlakeGuilloud: So instead of putting a ternary in the loadingRender function itself, you also have the option of writing it like this:

```javascript
        <Link to={route} className={classes} onClick={this.handleClick} disabled={disabled || loading} {...opts} title={this.props.title}>
          <span className="button__text-wrapper">{this.props.children}</span>
          {loading && this.loadingRender()}
        </Link>
```

and removing the `this.props.loading ? <content /> : null` in your function. I don't care either way, but thought I would point it out :)

zslabs: 👍 Yeah I like that! Will update.

zslabs:
Good catch on the description, I'll add that in.

PR #73, opened 2017-07-05
Zach/Modal + Cover append-to-body

zslabs:

### What should this PR do?
Moves the modal and cover render more inline with standard React components, but appends them to a separate subtree container using the [react-append-to-body](https://github.com/jpgorman/react-append-to-body) package.

**Note:** This PR cannot be merged as-is until there is some rework done to Modals and Covers in the app.

1. Test Modal and Cover state updates for form validation, etc.


BlakeGuilloud:
I think we may need to add this into the DropdownMultiSelect component. Leaving this here as a note for further investigation. 

```javascript
  componentWillReceiveProps(nextProps) {
    this.setState({
      activeKeys: nextProps.activeKeys,
    });
  }
```

PR #77, opened 2017-07-26
Blake/rhino switcher

BlakeGuilloud:

### What should this PR do?
This PR will add an additional prop `onChange` to CheckBox and RhinoSwitch components. The onChange prop should behave in the same way that existing form elements, such as Input and TextArea behave. This means we can use a single `handleChange = (name, value) =>` function for handling all of our form element changes.

This does not deprecate any existing functionality.

1. Confirm that switching a RhinoSwitch returns `name[string], value[bool]`.
2. Confirm that checking a Checkbox returns `name[string], `value[bool]`.


BlakeGuilloud:
If the two of you are OK with this PR, I may make additional changes to other form elements that do not currently return the name property onClick. I think we can all agree that they should? (i.e. Checkbox)

zslabs:
I like this! Do you think there's a lot of work on the RF side of things once this is merged in? I'm all for adding this in additional form elements and think you could just amend this PR with more.

zslabs:
Do you think the `onChange` property may work better?

BlakeGuilloud:
We can do onChange if you would prefer that, but to answer you first question- there would be no deprecation of functionality on RF, we could do a sweet over RF and change things like:

```javascript
<RhinoSwitch isChecked={desktopNotification} onClick={() => handleToggle('desktopNotification')} name="desktopNotification" />
```

into:

```javascript
handleToggle = (value, name) => this.setState({ [name]: value ]);

<RhinoSwitch isChecked={desktopNotification} onClick={handleToggle} name="desktopNotification" />
```

Now that I've written it, I kind of like your idea of putting in an onChange event for form element so it can follow the same pattern as Inputs and TextAreas. I'm going to move forward with that solution and revert some of the work I've done in this PR.

```javascript
handleChange = (name, value) => ({})
```

Thanks for the feedback!

PR #85, opened 2017-08-29
Missing file icon

zslabs:

screen shot 2017-08-29 at 2 15 51 pm

Will be used in attachments when images do not load properly.


zslabs:
@bworley29 @mcraiganthony Have any opinions on the icon used here? Just want to see if there was anything else we'd prefer.

mcraiganthony:
@zslabs Are we looking for a broken image icon or a missing file icon?

zslabs:
Spoke with @mcraiganthony on this, good to merge.

PR #86, opened 2017-09-01
Tooltips

zslabs:

### What should this PR do?
Converts our standard tooltips into GSAP-powered counterparts that provide a bit more flexibility for their content and display.

**Note:** When you see the event listeners, please don't yell at me. @BlakeGuilloud and I debugged this and React's synthetic events are not playing nice (firing twice). Happy to show if you'd like. You'll also see an update to the `<Icon />` component to not be a pure function. This was needed because I couldn't attach a `ref` to a pure function; something we'll need in the future with this component.

1. Tooltip wrapper should only be able to contain one, wrapped child
2. `content` prop can accept string, HTML, and/or other React components (see examples)
3. Tooltip DOM should be created and removed (reference DOM; at the bottom) which each mouseenter and mouseleave


src/scripts/components/Tooltip.jsx
```diff
@@ -0,0 +1,210 @@
...
+import { TimelineMax } from 'gsap';
+
+import { UtilitySystem } from '../UtilitySystem';
+
+class Tooltip extends React.Component {
+  // @NOTE Attaching event listeners here is not ideal, but `onMouseEnter` (and `onMouseOver`) is not reliably fired; as well as the context of the currentTarget being incorrect.
+  componentDidMount() {
+    const tooltipTrigger = this.getTooltipTrigger();
+
+    tooltipTrigger.addEventListener('mouseenter', (e) => {
```
alexgsummer: Don't forget to remove the event listener on `componentWillUnmount` for `mouseenter` and `mouseleave`.

zslabs: 👍 Thanks for the heads up, added that in.

zslabs:
@alexgsummer From the conversation we had yesterday, I added in touch support. I've added notes in the documentation, but could you give this a run-through to see if it works as you'd expect it to?

One item that won't be addressed at this time will be the automatic overflow to re-position based on available screen real-estate; that may come later and impact several components.

alexgsummer:
Looks good! Some things to consider, which may not fit your needs:

- Only allow one tooltip to be open on the page at a time. Opening one tooltip, auto closes any other ones.
- Attaching the close event to body, so you can click/touch anywhere to close the tooltip versus having to click back on the element itself.

zslabs:
Great idea on allowing only one to be open at a time! Added in some logic around that. For the second, I'll definitely keep that in mind as we determine how we're squeaking these into the app, could be something to revisit. Thanks again!

PR #89, opened 2017-09-20
Add tooltip delay

zslabs:

### What should this PR do?
Chatted with @mcraiganthony and this PR adds in the ability to delay a tooltip from showing up (as we may add it for hints on certain action items within the compose area), as well as also disabling it entirely on touch devices to not interfere with actionable items.

1. Test out delay feature. Can accept just the `delay` prop or a value `delay={2000}` (in milliseconds).


src/scripts/components/Tooltip.jsx
```diff
@@ -195,7 +171,19 @@ class Tooltip extends React.Component {
    * @return {void}
    */
   openTooltip(tooltip) {
-    tooltip.timeline.play();
+    const { delay } = this.props;
+
+    // Convert delay to a number
+    // Takes care of multiple PropTypes
+    const delayAmount = Number.isInteger(delay) ? Number(delay) : 1000;
```
alexgsummer: If `delay` is an integer, then you shouldn't need to convert it to a Number, right?

alexgsummer: Consider calling this variable `duration` or `delayDuration` versus `amount`. It's a personal preference, so no worries if you like `amount` better.

zslabs: So delay can either be a `bool` (which will default to that `1000`) or an integer, to override that default amount.

zslabs: 👍 yeah that makes sense

src/scripts/components/Tooltip.jsx
```diff
@@ -195,7 +171,19 @@ class Tooltip extends React.Component {
    * @return {void}
    */
   openTooltip(tooltip) {
-    tooltip.timeline.play();
+    const { delay } = this.props;
+
+    // Convert delay to a number
+    // Takes care of multiple PropTypes
+    const delayDuration = Number.isInteger(delay) ? Number(delay) : 1000;
```
alexgsummer: In response to your previous comment:
```
So delay can either be a bool (which will default to that 1000) or an integer, to override that default amount.
```
If I pass in `3000` for `delay`, `Number.isInteger(3000)` is `true`, so `Number(3000)` is returned. My question was around if it's an integer, it is already a number, so wrapping in `Number()` seems unnecessary.

If you pass in `true` or `false` for `delay`, `Number.isInteger(<bool>)` is `false`, and you'll get `1000` for both.

Is the intention that you can pass `false` and the delay is 0? You can pass `true` and the delay is defaulted to `1000`? Or you can pass a number and the delay will be that? I think the logic is off here or I may be missing something...

alexgsummer: Ok, my bad. I should have read further. You're checking below if delay is `true` or `false`. Ignore comment above, except maybe that you might not need the `Number()` call.

src/scripts/components/Tooltip.jsx
```diff
@@ -195,7 +171,19 @@ class Tooltip extends React.Component {
...
+
+    if (delay) {
```
alexgsummer: It might be clearer, if you put the `delayDuration` calculation inside this if statement since it's the only place you use it.

src/scripts/components/Tooltip.jsx
zslabs: 👍 Yep, you're right, we don't need `Number()` there.

src/scripts/components/Tooltip.jsx
zslabs: 👍

zslabs:
Ok, those comments have been addressed in the latest commit. Let me know if there's anything else.

PR #95, opened 2017-10-06
Modal/Cover logic cleanup

zslabs:

### What should this PR do?
Only shows modal (in DOM) based on separate state that is tied to the `open` prop and what it is currently equal to. This basically means we can write our Modals and Covers like any other component and know that there will not be any form ID collisions as we were once seeing.

**We can remove all of the logic we had around visibility for these components in RF. If you link over to it, you'll need to remove that logic as it can conflict with what this is doing.**

1. Loading doc page for Modals or Covers should not render either in DOM.
2. Clicking open will correctly add them to the DOM.
3. After the closing animation is complete, they will be removed from the DOM.
4. If you take out the state cleanup in the examples for `onReverseComplete` state should be preserved.


zslabs:
Updated to resolve conflicts

PR #100, opened 2017-10-23
React 16 (for when we want to tackle this again)

zslabs:


zslabs:
🎉 This one is back on! Talked with @benbruning and we'd like to get this in after Billing in RF is in.

PR #103, opened 2017-11-20
Date component

zslabs:

### What should this PR do?
Creates wrapper `<Date />` component that allows us to set sensible defaults for how we want to use the `<Datepicker />` library. Also not accepts `explanationMessage, label, validationMessage, required` props similar to a normal input. Users should also be able to type into the input by default, but we are going to handle the validation of that input on RF.


BlakeGuilloud:
Good stuff man.

PR #109, opened 2018-02-08
Zach/rg3/eslint

zslabs:

### What should this PR do?
Updates our `eslint-config-rhinogram` package to the latest versions and makes the appropriate changes throughout the app. There should be no functionality changes, only minor syntax updates. Most revolve around restructuring or keeping elements lined up as they span across multiple lines.

1. No brokey?


src/scripts/UtilitySystem/helpers/index.js
```diff
@@ -40,7 +40,7 @@ export const config = {
  * @return {void}
  */
 export function forEach(array, callback, scope) {
-  for (let i = 0, length = array.length; i < length; i += 1) {
+  for (let i = 0, { length } = array; i < length; i += 1) {
```
alexgsummer: I'm making the assumption that one of the eslint configs that was changed was one that always prefers destructuring...

IMO, this is strange syntax. Not in the sense that it doesn't make sense, but I've never seen the `length` property on an iterable destructured in this way before. @BlakeGuilloud @benbruning You guys have any preference on this?

zslabs: 👍 I'll update this.

PR #112, opened 2018-03-20
Cleave integration

zslabs:

### What should this PR do?
Integrates in the [Cleave](https://github.com/nosir/cleave.js/) library to our `<Input />` component. As mentioned during the standup today, provides input masking/validation for how users interact with our inputs. Typical use-cases would be dates, credit cards, and phone numbers -- but if you look through their options, you'll notice a number of things we can take advantage of. There were some unneeded checks around functions... being functions (since that's in our PropType validation anyway) so those were stripped out, but the input itself should function and accept everything we currently use it for, with the added `format` prop.

1. All existing callbacks and props work as expected in `<Input />` component.
2. Format options as available in https://github.com/nosir/cleave.js/blob/master/doc/options.md work as expected. __I did not include the phone validation library as part of the bundle in case projects we use it on don't require it, but if we are to take advantage of it in RF, it's part of what we can import into the project when in-use (as it comes with the package).__

One item to note is this only works on text, search, url, tel, and password input types, so anything else will throw an error. I made an issue of this on the repo https://github.com/nosir/cleave.js/issues/313 but decided against providing a fix on our end to see if they have a more elegant way of handling that in the library itself (I provided an example). Since we're the only consumers of the library, we can stick to those types pretty easily.

This PR does take away the datepicker (although it is potentially going away in the future) as there's some additional things that we'd still need to validate against on our end (minimum dates, making sure start/end times align, etc) in which I don't want to hold getting the infrastructure behind those future efforts in.


src/scripts/components/Input.jsx
```diff
@@ -40,36 +41,36 @@ class Input extends React.Component {
   _handleChange = (event) => {
     this.setState({ value: event.target.value });
 
-    if (this.props.onChange && typeof (this.props.onChange === 'function')) {
```
BlakeGuilloud: Thanks for getting rid of these.. In this case it was doing absolutely nothing since it was always checking "typeof true/false", which always returns a truthy value. They should have been written:
```
typeof this.props.onChange === 'function'
```
which would return true or false..

vs
```
typeof (this.props.onChange === 'function')
```
will return 'boolean'.

src/scripts/components/Input.jsx
```diff
@@ -40,36 +41,36 @@ class Input extends React.Component {
...
-      this.props.onChange(event.target.name, event.target.value.trimLeft());
+    if (this.props.onChange) {
+      this.props.onChange(event.target.name, event.target.value.trimLeft(), event.target.rawValue);
```
BlakeGuilloud: Do you think it's best to pass in `value` as the second argument instead of `rawValue`? The .value being the Cleave formatted string, whereas rawValue is what is actually in the input.

I'm fine with it either way, but it feels more backwards compatible passing in the rawValue as the second argument here. Just my opinion

zslabs: Ah yep, you're 100% right, good idea I'll update that.

src/scripts/docs/FormApp.jsx
```diff
@@ -182,6 +183,7 @@ const FormApp = () => (
 
     <section className="site-section">
       <h3 className="site-subheadline">Input</h3>
+      <p className="site-copy">We are using the <a href="http://nosir.github.io/cleave.js/">cleave</a> package to enhance out inputs with formatting/masking ability.</p>
```
alexgsummer: Spelling: "package to enhance **out** inputs"

zslabs:
Thanks for reviewing fellas -- noticed that if I apply `<Cleave />` to all elements, it strips spaces because it's assuming a special type of formatting. I'll need to make an update to make sure things work as expected for regular inputs.

src/scripts/docs/FormApp.jsx
zslabs: 🤦‍♂️

zslabs:
Alrighty - just pushed an update that should resolve that issue if you don't both mind taking another peak. Thanks again!

PR #114, opened 2018-04-05
Rg3 develop

zslabs:

### What should this PR do?
https://github.com/rhinogram/rhinostyle-legacy has been created to support RG2 moving forward, but this PR is to bring the latest RS up to speed in this repo --- also so we can finally update the demo site.


zslabs:
Going to do this a bit differently.

PR #115, opened 2018-04-19
Clean up form labels, validation, and explanation messages

zslabs:

### What should this PR do?
Reduces the redundancy of ☝️by using compositional components. 

1. Looks the same, acts the same, but much better.


src/scripts/components/RadioGroup.jsx
```diff
@@ -116,14 +90,15 @@ class RadioGroup extends React.Component {
...
 
   render() {
-    const { className } = this.props;
+    const { className, explanationMessage, label, required, validationMessage } = this.props;
     const classes = cx('form__group', className);
 
     return (
       <div className={classes}>
-        {this.showLabel()}
+        <FormLabel label={label} required={required} id="" />
```
BlakeGuilloud: Do we need this id=""?

zslabs: I have it as `required` in `<FormLabel />` because ideally when this is in-use, we'd have an associated input that we can hook into so when you click on the label; it focuses on it. I used `""` in this case since the labels here are more just a presentation layer since the radio group itself (and also in variable messages) doesn't have anything to focus on.

PR #119, opened 2018-05-03
Switch out component-playground for react-live

zslabs:

### What should this PR do?
Switches out `component-playground` for `react-live`; which is a library that gives us more flexibility on "how" the documentation is created, the ability to remove the Babel from our doc bundle (decreasing the size significantly). Also took this opportunity to clean-up how our documentation examples and scope are applied to the new `<Live />` component.

1. Examples should work as they do currently on the documentation site.


BlakeGuilloud:
Nice! ❤️

PR #121, opened 2018-07-30
Resource split (resource bottom)

zslabs:

### What should this PR do?
This adds a new sub-component to the `<Resource />` component, `<ResourceBottom />`. This provides us some additional flexibility for items placed beneath the main resource component (i.e. notes for the upcoming Connected Party adjustments).

1. Add `<ResourceBottom />` component as child to `<Resource />`.
2. Content should show up below main resource components (block-level).

screen shot 2018-07-30 at 2 43 20 pm


src/less/resources.less
```diff
@@ -245,6 +215,51 @@
       opacity: @global-opacity-disabled;
     }
   }
+
+  // We need to be a bit more relaxed with our borders since the container query watcher can get mixed in next to each `<Resource />` as state changes
```
alexgsummer: Spacing looks like it's off here for the comment and body of this rule.

src/scripts/components/Resource.jsx
```diff
@@ -2,9 +2,43 @@ import cx from 'classnames';
...
-import { ResourceRight, UtilitySystem } from '../components';
+import { ResourceBottom, ResourceRight, UtilitySystem } from '../components';
 
 class Resource extends React.Component {
+  getChildren = () => {
+    let returnChild = null;
+    const { children } = this.props;
+
+    return React.Children.map(children, (child) => {
+      if (!child) return false;
```
alexgsummer: I'm not sure if this check is needed (also below). The [documentation](https://reactjs.org/docs/react-api.html#reactchildren) says it will return `null` or `undefined` if children is `null` or `undefined`. If you do need this, maybe consider returning `null` like below rather than false.

src/less/resources.less
zslabs: Could I see a screen shot of what you're seeing? Is it looking different than in the app currently? I did need to change the wrapper, so definitely good to catch anything that looks off. This specific item is about "when" to show the border.

src/scripts/components/Resource.jsx
alexgsummer: I think you can just remove that line unless when you call that function you are specifically looking for boolean `false` or `null`. It seems like `React.Children.map` will return `undefined` / `null` for you and if all you care about is that it's "falsey"...

src/less/resources.less
alexgsummer: Sorry, ignore this comment. I thought the tilde character was something in Githubs diff instead of the CSS selector.

src/scripts/components/Resource.jsx
zslabs: I made the `null` change, but it looks like taking that check out completely busts the app. It could possibly be when we're looking for properties on children that may not be child "components" and not have sub-properties we're checking against, but looks like it may be a good idea to keep-in.

PR #125, opened 2018-09-25
Saubhagya/dropdown checkbox

saubhagya23:

### What should this PR do?
A new component **DropdownCheckbox** is added to rhinostyle under the category **Dropdowns**.

NA.

The new component is displayed as:
screen shot 2018-09-25 at 4 55 54 pm

when the checkbox is clicked, it looks like:
screen shot 2018-09-25 at 4 56 15 pm

### DIFF
--- a/src/scripts/components/DropdownCheckbox.jsx
+++ b/src/scripts/components/DropdownCheckbox.jsx
@@ -1,0 +1,268 @@
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import { Button, Checkbox, DropdownMenuItem, DropdownMenuItemWild, DropdownMenuScroll, DropdownFilter, DropdownWrapper, Icon } from '../components';
+
+class DropdownCheckbox extends React.Component {
+  state = {
+    activeKey: this.props.activeKey,
+    isOpen: false,
+    hasFilter: false,
+    isCheckbox: this.props.isCheckbox,
+    checked: this.props.isChecked,
+  };
+
+  componentWillMount() {
+    React.Children.map(this.props.children, (child) => {
+      if (!child) return;
+      if (child.type === DropdownFilter) {
+        this.setState({ hasFilter: true });
+      }
+    });
+  }
+
+  componentWillReceiveProps(newProps) {
+    if (newProps.activeKey !== this.props.activeKey) {
+      this.setState({
+        activeKey: newProps.activeKey,
+      });
+    }
+    if (newProps.isChecked !== this.props.isChecked) {
+      this.setState({
+        checked: newProps.isChecked,
+      });
+    }
+  }
+
+  getChildren = () => {
+    let returnChild = null;
+    const { children } = this.props;
+
+    return React.Children.map(children, (child) => {
+      if (!child) return null;
+
+      if (child.type === DropdownMenuItem) {
+        const onClick = () => {
+          if (child.props.id) {
+            if (this.props.onSelect) {
+              this.updateActiveKey(child.props.id);
+              this.props.onSelect(child.props.id);
+            } else {
+              this.updateActiveKey(child.props.id);
+            }
+          }
+
+          if (child.props.onClick) {
+            child.props.onClick();
+          }
+
+          this.handleToggle();
+        };
+
+        returnChild = React.cloneElement(child, {
+          onClick,
+          active: this.state.activeKey && (child.props.id === this.state.activeKey) ? true : false, // eslint-disable-line
+        });
+      } else if (child.type === DropdownFilter) {
+        returnChild = React.cloneElement(child, {
+          onSelect: this.props.onSelect,
+          handleToggle: this.handleToggle,
+          activeKey: this.state.activeKey,
+          icon: this.props.icon,
+          updateActiveKey: this.updateActiveKey,
+        });
+      } else if (child.type === DropdownMenuItemWild) {
+        returnChild = React.cloneElement(child, {
+          handleToggle: this.handleToggle,
+        });
+      } else {
+        returnChild = child;
+      }
+
+      return returnChild;
+    });
+  }
+
+  handleToggle = () => {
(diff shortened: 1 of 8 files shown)


src/scripts/docs/examples/DropdownCheckbox.example.txt
```diff
@@ -0,0 +1,44 @@
...
+      this.setState({
+        labelValueAssociated: '',
+        checkboxClassName: '',
+      });
+    } else {
+      this.setState({
+        labelValueAssociated: 'All Selected',
+      }); 
+    }
+    this.setState({ checked: !this.state.checked });
```
mahimaag: Please use prevState here to get the previous value of checked.

src/scripts/docs/examples/DropdownCheckbox.example.txt
```diff
@@ -0,0 +1,44 @@
...
+  render() {
+    const itemArray = [{ id: 1, label: 'Select All', description: 'All Selected', isChecked: true, associatedLabel: 'All Selected', checkboxClassName: '' },
+    { id: 2, label: 'None', description: 'None', isChecked: false, checkboxClassName: '' },
+    { id: 3, label: 'Following', description: 'Following', isChecked: true, associatedLabel: '2 Selected', checkboxClassName: 'partially-checked' }];
+  
+    return (
+      <UtilityInlineGrid>
+        <DropdownCheckbox isCheckbox type="checkbox" isChecked={this.state.checked} showAssociatedLabel labelValueAssociated={this.state.labelValueAssociated} onChange={this.handleChecked} checkboxClassName={this.state.checkboxClassName}>
+          {itemArray.map(item => <DropdownMenuItem key={item.id} id={item.id} label={item.label} onClick={() => this.handleSelect(item)} />)}
+        </DropdownCheckbox>
```
mahimaag: Use indentation formatting here.

src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -0,0 +1,268 @@
...
+      }
+
+      return returnChild;
+    });
+  }
+
+  handleToggle = () => {
+    const $dropdown = ReactDOM.findDOMNode(this.dropdown);
+
+    if (this.state.isOpen) {
```
abhicool957: It would be good to destructure here like below:
`const { isOpen } = this.state; `
and then use `isOpen` anywhere you want.

src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -0,0 +1,268 @@
...
+      // Close dropdown
+      $dropdown.timeline.reverse();
+    } else {
+      // Open dropdown
+      $dropdown.timeline.play();
+    }
+
+    // Update state
+    this.setState({
+      isOpen: !this.state.isOpen,
```
abhicool957: We should use here `preState` rather than negate the state property.

src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        return <span className="dropdown__toggle__text">{selectedLabel || label}</span>;
+      }
+      return false;
+    };
+    const showCheckbox = () => {
+      const { isCheckbox } = this.state;
+      if (isCheckbox) {
+        return <Checkbox name="test" label=" " isChecked={this.state.checked} onChange={this.props.onChange} className={checkboxClassName} />;
```
abhicool957: Please use destructuring here also for `this.state.checked` and `this.props.onChange`

src/scripts/components/DropdownCheckbox.jsx
abhicool957: use `preState` here also. You can refer this from Review document.

src/scripts/docs/examples/DropdownCheckbox.example.txt
```diff
@@ -35,8 +35,22 @@ class ComponentExample extends React.Component {
   
     return (
       <UtilityInlineGrid>
-        <DropdownCheckbox isCheckbox type="checkbox" isChecked={this.state.checked} showAssociatedLabel labelValueAssociated={this.state.labelValueAssociated} onChange={this.handleChecked} checkboxClassName={this.state.checkboxClassName}>
-          {itemArray.map(item => <DropdownMenuItem key={item.id} id={item.id} label={item.label} onClick={() => this.handleSelect(item)} />)}
+        <DropdownCheckbox 
+          isCheckbox type="checkbox"
```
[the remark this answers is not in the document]
saubhagya23: No, they're not. I missed it. Thanks for pointing that one out.

saubhagya23:
@abhicool957 The changes you suggested have been made. Please review.

src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -0,0 +1,268 @@
...
+    });
+  }
+
+  componentWillReceiveProps(newProps) {
+    if (newProps.activeKey !== this.props.activeKey) {
+      this.setState({
+        activeKey: newProps.activeKey,
+      });
+    }
+    if (newProps.isChecked !== this.props.isChecked) {
```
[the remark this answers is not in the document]
saubhagya23: @abhicool957  Since **isChecked** contains a boolean value, we do not need to check it's presence here. If we do apply the check to verify if the value is present, it would only work when value will be true.

src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -0,0 +1,268 @@
...
+    React.Children.map(this.props.children, (child) => {
+      if (!child) return;
+      if (child.type === DropdownFilter) {
+        this.setState({ hasFilter: true });
+      }
+    });
+  }
+
+  componentWillReceiveProps(newProps) {
+    if (newProps.activeKey !== this.props.activeKey) {
```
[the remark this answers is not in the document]
saubhagya23: @abhicool957 **activeKey** is used to represent the current selected item. This condition is used to update the current selected item. If we don't provide **activeKey** property, current selected item would not be updated.

PR #127, opened 2018-09-27
RHIN-983: changes made to resourceIntro component to incorporate checkbox.

saubhagya23:

### What should this PR do?
When **checkbox** is added as props to **resourceIntro** in rhinostyle, a checkbox should be displayed.
https://rhinogram.atlassian.net/browse/RHIN-983

When **checkbox** is added as props to **resourceIntro** in rhinostyle, a checkbox should be displayed.

when **checkbox is added as props** , it appears as below:

screen shot 2018-09-27 at 12 41 46 pm

### DIFF
--- a/src/scripts/components/ResourceIntro.jsx
+++ b/src/scripts/components/ResourceIntro.jsx
@@ -29,7 +29,7 @@ class ResourceIntro extends React.Component {
                 label={checkbox.label}
                 name={checkbox.name}
                 isChecked={checkbox.isChecked}
-                onClick={checkbox.onClick}
+                onChange={checkbox.onChange}
               />
             }
             <Button reset onClick={this.handleIconClick}>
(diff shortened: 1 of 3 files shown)


src/scripts/components/ResourceIntro.jsx
```diff
@@ -16,19 +16,41 @@ class ResourceIntro extends React.Component {
...
-          <Button reset onClick={this.handleIconClick}>
-            <Icon bump={icon.bump} icon={icon.icon} />
-          </Button>
+          <div className="u-flex">
+            {checkbox &&
+              <Checkbox
+                label={checkbox.label}
+                name={checkbox.name}
+                isChecked={checkbox.isChecked}
+                onClick={checkbox.onClick}
```
mahimaag: Change onClick to onChange. Apply everywhere

src/scripts/docs/examples/Resource.example.txt
```diff
@@ -64,7 +75,7 @@ class ComponentExample extends React.Component {
             </ResourceRight>
           </Resource>
         </ResourceGroup>
-        <ResourceGroup separator="Another example" interfaceMode="radio">
+        <ResourceGroup separator="Another example" interfaceMode="checkbox">
```
mahimaag: Are we using checkbox instead of radio button?

saubhagya23:
@mahimaag The changes you suggested has been made.

PR #129, opened 2018-10-05
testing new docs folder

adlondon:

### What should this PR do?
Created docs directory


src/pages/animations.html
```diff
@@ -4,7 +4,7 @@
 
 
 <section class="site-section">
-  <h3 class="site-subheadline">Animation Types</h3>
+  <h3 class="site-subheadline">Animation Tips</h3>
```
alexgsummer: Did you mean to change this to Tips?

package.json
```diff
@@ -33,10 +33,9 @@
     "cssnano": "3.10.0",
     "del": "3.0.0",
     "eslint-config-rhinogram": "0.0.10",
-    "gulp": "3.9.1",
+    "gulp": "^3.9.1",
```
alexgsummer: We prefer to lock down our dependencies to specific versions, so here remove the carets (^).

adlondon: Haha, was testing to make sure changes were coming through. Ill fix that right quick

src/pages/animations.html
adlondon: Haha, was testing to make sure changes were coming through. Ill fix that right quick

PR #132, opened 2018-10-10
Andrew/docs

adlondon:

### What should this PR do?
Removed carrats and updated README


adlondon:
@benbruning @alexgsummer had to add dist command to the script watch

PR #133, opened 2018-10-11
RHIN-431: changes made to checkbox, resourceIntro and dropdownCheckbox

saubhagya23:

### What should this PR do?
When selecting/unselecting the checkbox, the dropdown list gets selected and unselected accordingly.
The click event of resource and the checkbox within a resource separated.

When selecting/unselecting the checkbox, the dropdown list gets selected and unselected accordingly.
The click event of resource and the checkbox within a resource separated.


src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -36,6 +36,16 @@ class DropdownCheckbox extends React.Component {
...
+  onCheckboxChange = () => {
+    const { checked } = this.state;
+    if (checked) {
+      this.setState({ activeKey: 1 });
+    } else {
+      this.setState({ activeKey: 0 });
+    }
+    this.props.onChange();
+  }
+
```
mahimaag: We can pass active key as props from where we are calling this component. No meed to set activeKey here.

src/less/resources.less
```diff
@@ -273,6 +273,16 @@
   &.has-avatar {
     align-items: center;
   }
+
+  .rhinobox__input {
+    position: absolute !important;
+    -webkit-appearance: none;
+    height: 2rem !important;
+    width: 2rem !important;
+    z-index: 99999;
```
alexgsummer: It looks like this project is using `z-index: 1000` to indicate the highest z-index value.

mahimaag:
@alexgsummer, suggested changes has been done. Please review.

PR #137, opened 2018-10-26
Add inputBlur props to dropdown to enable blur on inputs inside dropdown menu.

rosh-ttn:
This change has been done with respect to the subtask RHIN-1180.
Jira Link:  https://rhinogram.atlassian.net/browse/RHIN-1180

### What should this PR do?
Set and unset focus on inputs inside dropdown depending on inputBlur props passed to dropdown.

1. List


src/scripts/components/Dropdown.jsx
```diff
@@ -134,7 +134,7 @@ class Dropdown extends React.Component {
   }
 
   render() {
-    const { avatar, block, className, disabled, disableScroll, hideCaret, label, icon, lockLabel, position, reset, size, title, type, wide, onStart, onComplete, onReverseStart, onReverseComplete, showOverflow, wrapperClassName } = this.props;
+    const { avatar, inputBlur, block, className, disabled, disableScroll, hideCaret, label, icon, lockLabel, position, reset, size, title, type, wide, onStart, onComplete, onReverseStart, onReverseComplete, showOverflow, wrapperClassName } = this.props;
```
alexgsummer: Instead of `inputBlur`, use a variable similar to the html5 `autoFocus` property. Maybe `autoFocusInput` as the property name. Typically, blur is associated with losing focus and here the input doesn't have focus yet so there is nothing to blur.

Once you change it to `autoFocusInput`, you will need to default the value to `true` so that we don't affect other places in the app that depend on this behavior.

rosh-ttn: I have updated it to a default value of true and changed name of prop. But, we will have to make sure it do not affects the other parts of rhinogram frontend as we now need to pass autoFocusInput props to dropdown wherever we required a focus on input as per the UI of our existing application. @alexgsummer

rosh-ttn: @alexgsummer The default behaviour in the app is that the input field inside the dropdown is in the focus state, as it was earlier. But, in that case when I do not want focus on the inputs inside the dropdown menu (as in the issue with datepickers inside the dropdown menu), I am passing a property named as inputBlur to the dropdown, which blurs  the focus set on the inputs. Hence, the other parts of the app are not affected by this update.

src/scripts/components/DropdownWrapper.jsx
```diff
@@ -35,7 +35,7 @@ class DropdownWrapper extends React.Component {
         const $input = $dropdown.querySelector('input[type="text"]');
 
         // If dropdown contains input (for filtering), focus on that
-        if ($input) {
+        if ($input && !this.props.inputBlur) {
           $input.focus();
```
rosh-ttn: @alexgsummer So, whenever I pass inputBlur to dropdown, the focus property is set only when the value of inputBlur is false. In cases when I am not passing inputBlur props, the behaviour of the inputs remains unaffected.

src/scripts/components/Dropdown.jsx
rosh-ttn: @alexgsummer I have renamed the prop and I am passing autoFocusInput to false in the dropdown component whenever I need to prevent focus on the inputs. Also whenever we are not passing this prop, the behaviour of the app is not affected. Please review.

src/scripts/components/DropdownWrapper.jsx
```diff
@@ -34,8 +34,8 @@ class DropdownWrapper extends React.Component {
       onComplete: () => {
         const $input = $dropdown.querySelector('input[type="text"]');
 
-        // If dropdown contains input (for filtering), focus on that
-        if ($input) {
+        // If dropdown contains input (for filtering), focus on that if autoFocusInput props is true or undefined
+        if ($input && (this.props.autoFocusInput === undefined || this.props.autoFocusInput !== false)) {
```
alexgsummer: Use React's `defaultProps` to default `autoFocusInput` value to `true`.

rosh-ttn: @alexgsummer Updated as suggested. Thanks this was useful :+1:

PR #140, opened 2018-10-31
Andrew/update variable messages

adlondon:

### What should this PR do?
Fixes casing for variable names and adds the ability for them to be deleted via the keyboard


src/scripts/components/ToggleButton.jsx
```diff
@@ -5,7 +5,7 @@ import React, { Component } from 'react';
 class ToggleButton extends Component {
   handleClick = () => {
     if (this.props.onClick) {
-      this.props.onClick(this.props.variable.value, this.props.variable.id);
+      this.props.onClick(this.props.variable.variable, this.props.variable.id);
```
benbruning: Is this intended?

src/scripts/components/VariableMessage.jsx
```diff
@@ -82,14 +82,24 @@ class VariableMessage extends React.Component {
...
-
-    // Replace text contents of variable to hide squigglies
-    const regexSquiggles = /{(.*)?}/g;
-    value = value.replace(regexSquiggles, '<i>{</i>$1<i>}</i>');
-
-    // Replace text contents of variable to hide underscore
-    const regexUnderscores = /_/g;
-    value = value.replace(regexUnderscores, '<b>_</b>');
+    // We are treating these as a special case because formatting for the "to" needs to me lowercase
+    if (text === '{REPLY_‘1’_to_CONFIRM}' || text === '{REPLY_‘2’_to_CANCEL}') {
```
benbruning: I wonder if @alexgsummer has any suggestions for this, I don't really love the idea of having hard coded logic in rhinostyle for specific variable messages.

src/scripts/components/ToggleButton.jsx
adlondon: it is. i was doing some goofiness before that i really didnt neeed to do.

alexgsummer: Can you elaborate on this?

src/scripts/components/VariableMessage.jsx
```diff
@@ -210,6 +212,28 @@ class VariableMessage extends React.Component {
     }
   }
 
+  handleKeyUp = (k) => {
+    // check if delete key or backspace is pressed to see if a variable was removed
+    if (k.which === 8 || k.which === 46) {
```
alexgsummer: Consider making the key codes here constants.

src/scripts/components/VariableMessage.jsx
```diff
@@ -142,16 +141,20 @@ class VariableMessage extends React.Component {
...
       const { variable } = value;
-      const isVariablePresent = split.includes(variable);
-
+      const isVariablePresent = lowercaseSplit.includes(variable.toLowerCase());
+      console.log(variable.toLowerCase(), isVariablePresent);
       // See if we've found one in our `initialValue`
       if (isVariablePresent) {
-        const variableIdx = split.indexOf(variable);
+        // If so, transform the variable into HTML
+        const variableIdx = lowercaseSplit.indexOf(variable.toLowerCase());
```
alexgsummer: Consider not abbreviating `Index`. I see it may have been used previously, but try to convert them or change them as you go.

src/scripts/components/ToggleButton.jsx
adlondon: It was leftover from changes i was trying to make / test in rhinofront  by inserting the variable value (ie First Name) instead of the actual variable ({first_name}). I thought that by inserting the value I would be able to avoid having to transform the actual variable, but that ended up not being the case. Here i'm just changing it back to how its supposed to be

adlondon:
@alexgsummer updates made

src/scripts/components/ToggleButton.jsx
alexgsummer: I see. `variable.variable` is kinda weird. Ideally the names would be different to help differentiate them. If it was already like that, then I'm fine with it.

src/scripts/components/VariableMessage.jsx
```diff
@@ -210,6 +210,30 @@ class VariableMessage extends React.Component {
     }
   }
 
+  handleKeyUp = (k) => {
+    const backspaceKeyPressed = k.which === 8;
```
alexgsummer: Sorry, I meant the key code value. So in the constants file you might create a section for key codes and create a constant named `BACKSPACE_KEY`:
```
export const BACKSPACE_KEY = 8;
```

The idea is to not have any "magic numbers" in code. So when reading this code it reads:
```
if (k.which === BACKSPACE_KEY) [...]
```

adlondon:
@alexgsummer got it, changes made

src/scripts/components/VariableMessage.jsx
```diff
@@ -2,8 +2,18 @@ import cx from 'classnames';
...
+import { Button,
+  FormLabel,
+  FormExplanationMessage,
+  FormValidationMessage,
+  Message,
+  ToggleButton,
+  UtilitySystem,
+} from '../components';
+
+const BACKSPACE_KEY = 8;
```
benbruning: I know I approved already, but responding to the point @alexgsummer brought up.  If we are making this change, does it make sense to create a constants file in rhinostyle? I could see these being sharable amongst other components in the future.

adlondon: Makes sense to me. `constants.js` file added to `/src`. Let me know if you'd like it to live somewhere else.

PR #148, opened 2018-11-15
RHIN-1274 Rhinostyle component for multiselect dropdown

yogitattn:
https://rhinogram.atlassian.net/browse/RHIN-1274

### What should this PR do?
This PR is for Multiselect new design dropdown component for Audit filters

NA


src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,246 @@
...
+} from '../components';
+
+class DropdownMultiSelectAdvanced extends React.Component {
+  state = {
+    searchText: '',
+    viewAllUsers: true,
+  }
+
+  sortMemberSearchIds = (a, b) => {
+    // Sort the rest by lastName
```
mahimaag: Sorting is performing on first name here. Change comment.

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,246 @@
...
+    }
+    this.props.handleUpdateSelectedIds(selectedIds, selectedUsers);
+  }
+
+  handleToggle = () => {
+    if (this.state.viewAllUsers) {
+      this.setState({ searchText: '' });
+      this.props.fetchUsersSearch('');
+    }
+    this.setState({ viewAllUsers: !this.state.viewAllUsers });
```
mahimaag: Use prevState here for toggelling value of viewAllUsers.

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,246 @@
...
+  UtilitySystem,
+  Dropdown,
+  UtilityInlineGrid,
+  ResourceRight,
+} from '../components';
+
+class DropdownMultiSelectAdvanced extends React.Component {
+  state = {
+    searchText: '',
+    viewAllUsers: true,
```
mahimaag: Add 'is' before Boolean type variable name.

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,246 @@
...
+  render() {
+    const { userSearchLoading, openPanel, dropdownLabel, selectedUserIds, filterName } = this.props;
+    const usersIds = [...this.props.usersIds];
+    const title = `View Selected (${this.props.selectedUserIds.length})`;
+    const searchTitle = `Search ${filterName}`;
+    let returnValue = '';
+    if (openPanel) {
+      returnValue = (
+        this.state.viewAllUsers ?
+          (
```
mahimaag: Move parenthesis up

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,246 @@
...
+                      </Button>
+                      <Button
+                        size="small"
+                        type="link"
+                        onClick={this.handleToggle}
+                        disabled={this.props.selectedUserIds.length === 0}
+                        title={title}
+                      >{title}
+                      </Button>
+                    </UtilityInlineGrid>)
```
mahimaag: I am seeing this code is repeated on more that one place. Wrap this code in a separate function if possible and use that function.

src/scripts/components/DropdownMultiSelectAdvanced.jsx
juanfabrega: Hey @yogitattn and @mahimaag, one thing I noticed is we have member/user specific logic in here. I know that comes from the acceptance criteria. Is there a way to make this more generic to create a re-usable component? While we're only using this for a user dropdown at the moment, it's likely to be used in other cases moving forward. The component should be able to handle a variety of items and not be locked to members and last name sorting. The member-specific stuff should be passed to the component as context/props in rhinofront.

Let me know if there's anything I can do to help clarify. Thanks!

yogitattn:
@mahimaag I have done all the changes as per your feedback

src/scripts/components/DropdownMultiSelectAdvanced.jsx
juanfabrega: @alexgsummer @benbruning what do y'all think of the approach here. I don't have enough context on this specific feature, but the original thought when I wrote the comment is that the component could be used for other lists besides users in the future (groups, channels, etc). 

If that's not the case then it is likely ok to keep as is, but if there is potential for re-use of the UI pattern, then I would recommend handling things like sorting logic in rhinofront and removing any references to users/members and replacing with list items.

yogitattn: @juanfabrega Now I have amde some changes, removed the sorting logic and moved it to rhinofront side.

benbruning: I agree with @juanfabrega, while this may only currently be used for Users specifically right now, rhinostyle components should be generic and be able to be used for other types of data.  If it needs to be so specific to users, it probably shouldn't be a rhinostyle component.

alexgsummer: Agreed. This component shouldn't use `users` for any variable names and remain generic. 

Functions, like `formatAvatarName` probably shouldn't be in RhinoStyle. The consumer of this component will either need to pass in a formatter function or pass in data that can be consumed and displayed "as-is".

yogitattn: @juanfabrega @alexgsummer @benbruning I have made all changes regarding making this component fully generic. Now It can be used for another type of list also, In case of a user, We just need to pass a prop 'type' as 'member' to render avatar, but for other types, only list will be displayed with titles only. I have also made  all variables and functions names generic without any 'user' keyword and have removed this 'formatAvatarName' function from rhinostyle . Please review once.

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,235 @@
...
+  );
+
+  render() {
+    const { itemSearchLoading, dropdownLabel, selectedItemsIds, filterName } = this.props;
+    const itemsIds = [...this.props.itemsIds];
+    const searchTitle = `Search ${filterName}`;
+    let returnValue = '';
+    returnValue = (
+      this.state.isViewAllItems ?
+        (
```
alexgsummer: Bring this open parens up one line.

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,235 @@
...
+          <Dropdown wide autoFocusInput={false} label={dropdownLabel} onClick={this.clearSearch} type="outline-primary" disableScroll>
+            <div className="dropdown__menu__container">
+              <div className="search__group">
+                {selectedItemsIds.length > 0 ? (
+                  <UtilityInlineGrid className="u-flex u-flex-justify-between u-m-t-small u-text-small">
+                    {this.renderClearButton()}
+                    {this.renderViewSelected()}
+                  </UtilityInlineGrid>)
+                  :
```
alexgsummer: Use the format:
```
) : (
```

src/scripts/docs/examples/DropdownMultiSelectAdvanced.example.txt
```diff
@@ -0,0 +1,82 @@
...
+        },    
+    itemsIds: [],
+    itemsCopy: {},    
+  }
+
+  componentWillMount(){
+    this.setState({ itemsIds: Object.keys(this.state.items), itemsCopy: this.state.items });
+  }
+
+  fetchAllItems = (searchText) => {
```
alexgsummer: Try not to use `fetch` when you're not making a network request or getting data from an external resource.

src/scripts/docs/examples/DropdownMultiSelectAdvanced.example.txt
```diff
@@ -0,0 +1,82 @@
...
+  //This code is written for a case, If we are filtering from the store data only and not from api, Otherwise we will only call api
```
alexgsummer: Please move this comment above the function declaration. It says it will call API, but I'm not seeing where that is done. Consider making this comment more clear.

src/scripts/docs/examples/DropdownMultiSelectAdvanced.example.txt
yogitattn: @alexgsummer In case of our members and contacts filters dropdown in rhinofront, I will be fetching data from api, that's why I used it.

src/scripts/docs/examples/DropdownMultiSelectAdvanced.example.txt
yogitattn: @alexgsummer Ok I will make it more clear in code also, From this comment I want to say that, I am not making any api call in this example, I am filtering the data from the stored state, but in case of members and  contacts filter of audit log, I will be calling api. So, this code can be used for both cases, but in the example i have used only one case.

yogitattn:
@alexgsummer I have done all changes as you have suggested. Please review

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -0,0 +1,233 @@
...
+          <div className="dropdown__menu__container">
+            {itemsIds.length > 0 ? (
+              <Scrollbars className="resource-group__scroll" autoHeight autoHeightMax={UtilitySystem.config.resourceSizes.large}>
+                <ResourceGroup interfaceMode="checkbox">
+                  {itemsIds.map(this.renderList)}
+                </ResourceGroup>
+              </Scrollbars>
+            ) :
+              this.renderSearchHelp(itemsIds, itemSearchLoading)
+              }
```
alexgsummer: I don't think this bracket lines up with the opening bracket.

yogitattn: @alexgsummer Ok I will change it before merge

PR #151, opened 2018-11-15
RHIN-1290: Handling on check on dropdown checkbox in Firefox

rosh-ttn:

### What should this PR do?
Enable checked state on checkbox on click inside checkbox and dropdown toggle on clicking on caret
1. List


src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -244,10 +239,10 @@ class DropdownCheckbox extends React.Component {
           {showCheckbox()}
           {hideCaret || (icon && !label && !selectedLabel) ?
             null :
-            <Icon size="small" icon="caret-down" className="dropdown__toggle__caret" />
+            <Icon size="small" icon="caret-down" className="dropdown__toggle__caret dropdown__caret--auto" />
```
juanfabrega: How does this work with multiple classes on the className attribute? Do these need to be generated with cx method?

rosh-ttn: There was an alignment issue with svg's inside the dropdown button. I fixed this with another css solution and removed the class dropdown__caret--auto from the code. @juanfabrega

src/scripts/components/DropdownCheckbox.jsx
```diff
@@ -244,10 +239,10 @@ class DropdownCheckbox extends React.Component {
           {showCheckbox()}
```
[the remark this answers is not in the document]
rosh-ttn: Updated as suggested. Thanks :+1: @juanfabrega

PR #152, opened 2018-11-15
Removing dist directory git tracking and adding build hook on install

alexgsummer:
### What should this PR do?
This removes the dist directory and contents from git tracking and adds a `prepare` hook which will use `gulp` (from node_modules of this project) to build the project when you `yarn install`.

You should not need gulp installed globally for this hook to work (when building in `rhinofront` for instance).


adlondon:
Do you think we could add a preinstall script to the package.json that runs `yarn add gulp -g` or something similar to ensure that doesnt give anyone errors, or is that overkill?

PR #159, opened 2018-11-26
Customized legends for doughnut charts as per design.

rosh-ttn:

### What should this PR do?
[1-3 sentences]

Not Applicable

dnc


src/less/charts.less
```diff
@@ -61,4 +61,50 @@
...
+  .chart__doughnut--info {
+    div {
+      margin-top: @global-whitespace-small!important;
+    }
+
+    .list--title {
+      font-size: @global-font-size-xsmall!important;
+    }
+  }
+}
```
mahimaag: Use BEM convention here.

komalsingh54: done

juanfabrega:
@komalsingh54 Approved, but can we look at an alternative to componentWillMount? We're still using that method in older areas, but it is deprecated so we want to avoid using it on any new code. 

This code should move to componentDidMount or the component constructor.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
-const renderNoData = () => (
-  <div className="chart__without-data">
-      Sorry! There&apos;s nothing to show.
-    <p>
-      Once data is available for this section, it will appear here.
-    </p>
-  </div>
-);
+  legendCallback = function(chart) {  // eslint-disable-line
+    const { labels } = chart.data;
```
alexgsummer: Add `datasets` here so you can use it below instead of `chart.data.datasets`.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
+    const { data } = chart.data.datasets[0];
+    const { backgroundColor } = chart.data.datasets[0];
```
alexgsummer: Add `backgroundColor` above since you're already destructuring `chart.data.datasets[0]`.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
+    return text.join('');
+  };
 
-const Chart = (props) => {
-  const { ...opts } = props;
-  let isChartData = false;
-  if (typeof opts.data === 'undefined' || opts.data === null || Object.keys(opts.data).length === 0 || Object.keys(opts.data.datasets[0].data).length === 0) {
-    isChartData = true;
+  componentWillMount() {
+    const { ...opts } = this.props;
```
alexgsummer: Looks like you're only using `this.props.data`, so only destructure that property.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
-              <span className="subtitle--muted">
-                {opts.subHeader}
-              </span>
-            )}
+
+  componentDidMount() {
+    this.forceUpdate();
+  }
+
+  renderChart = (opts) => {
```
alexgsummer: Don't abbreviate. `Options` could be more specific, especially since below you're doing `opts.options`.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
+    switch (opts.type.toLowerCase()) {
+      case 'bar':
+        return (<Bar {...opts} />);
+      case 'doughnut':
+        opts.options.legendCallback = this.legendCallback; // eslint-disable-line
```
alexgsummer: If you need to disable an eslint rule, add the specific rule so it doesn't disable all rules.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
-      return (<Doughnut {...opts} />);
-    case 'line':
-      return (<Line {...opts} />);
-    default:
-      return (<Bar {...opts} />);
-  }
-};
+class Chart extends React.Component {
+  state = {
+    isChartData: false,
```
alexgsummer: `isChartData`? Does this indicate whether the data is available? If so, consider `hasChartData`.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
-                {opts.subHeader}
-              </span>
-            )}
+
+  componentDidMount() {
+    this.forceUpdate();
+  }
+
+  renderChart = (opts) => {
+    switch (opts.type.toLowerCase()) {
```
alexgsummer: Prefer regular IF statements, when you don't need the fall-through mechanism of switch.

src/scripts/components/Chart.jsx
```diff
@@ -1,66 +1,104 @@
...
+      Sorry! There&apos;s nothing to show.
+      <p>
+        Once data is available for this section, it will appear here.
+      </p>
     </div>
   );
-};
 
+  render() {
+    const { ...opts } = this.props;
```
alexgsummer: Destructure the specific properties you need.

src/scripts/components/Chart.jsx
komalsingh54: done

src/scripts/components/Chart.jsx
komalsingh54: done

src/scripts/components/Chart.jsx
komalsingh54: done

src/scripts/components/Chart.jsx
komalsingh54: done

src/scripts/components/Chart.jsx
komalsingh54: done

src/scripts/components/Chart.jsx
komalsingh54: done

src/scripts/components/Chart.jsx
komalsingh54: done

src/scripts/components/Chart.jsx
komalsingh54: I am using complete object here.

komalsingh54:
@alexgsummer i have made the feedback changes

src/scripts/components/Chart.jsx
alexgsummer: Here I didn't mean to use `Options`. I meant try to come up with a better variable name because `opts.options` is confusing / vague.

komalsingh54: Yup, rename the variable name. thanks

PR #160, opened 2018-11-27
Minor changes in example of multiselect dropdown

yogitattn:
### What should this PR do?
Minor changes in multiselect dropdown component example

NA


yogitattn:
@alexgsummer @benbruning @juanfabrega There was a minor change in multiselect dropdown example, before, i was using firstname and lastname in component, but in later commits I made component more generic and used 'title' key instead. This minor change was also needed in example. So, I am merging these changes.

PR #162, opened 2018-11-30
Update smart table with fixed first column for mobile views

rosh-ttn:

### What should this PR do?
First column of the smart table should be fixed while keeping rest of the table body horizontally scrollable.

Not Applicable


src/less/tables.less
```diff
@@ -698,3 +698,59 @@
   }
 }
 
+@media (max-width: @global-screen-small-max) {
+
```
alexgsummer: Remove empty line here.

src/less/tables.less
```diff
@@ -698,3 +698,59 @@
...
+  .ReactTable {
```
alexgsummer: Are we using capital letters for class names anywhere else?

src/less/tables.less
```diff
@@ -698,3 +698,59 @@
...
+    .rt-table {
```
juanfabrega: Hey @rosh-ttn it feels like we're not using much of the established style variables that we have and have instead targeted some very specific measurements, with lots of !important attributes that could affect CSS hierarchy in the long-term. 

I am not super familiar with the front-end design for this but I will pull the branch and provide some more concrete comments as soon as I can so we can refactor this.

src/less/tables.less
rosh-ttn: Actually, these classes are generated by React Table and as per design, I wanted to overwrite few css rules for these. @alexgsummer

alexgsummer:
Removing myself as PR reviewer on this one. I'm good when @juanfabrega approves.

src/less/tables.less
juanfabrega: Hey @rosh-ttn I'm meeting with benwo on this today just to understand the functionality and I will run him through the progress using this branch. Hoping to have any remaining feedback to you this morning.

src/less/tables.less
```diff
@@ -698,3 +698,59 @@
...
+        width: 100px !important;
+        left: 4px;
+        top: auto;
+        background-color: #fff;
+      }
+
+      .rt-tr {
+        &.-odd {
+          .rt-td:nth-child(1) {
+            background-color: #00000008 !important;
```
juanfabrega: Too many digits on this hex value. Also can we use a variable? Any of our variables match the color we need here?

src/less/tables.less
```diff
@@ -698,3 +698,59 @@
...
+      }
+
+      .rt-th {
+        span {
+          margin-left: 1px;
+        }
+      }
+
+      .rt-td,
+      .rt-th {
```
juanfabrega: Can we use some of our global spacing variables rather than new pixel values? If not, just let me know, but want to make sure we keep everything consistent in the UI.

src/less/tables.less
rosh-ttn: @juanfabrega I found out the solution for minimizing the use of !important in the code. Will be pushing the updated code. Also, I will replace this with the existing variable we have in variables list.

src/less/tables.less
rosh-ttn: Will try to replace them with the closest match from the variable list. +1 @juanfabrega

src/less/tables.less
juanfabrega: Awesome @rosh-ttnm, thanks! Let me know when it is ready to review. 

As I said earlier, if this is blocking progress, we are ok to move ahead and I have opened a card for future refactoring in JIRA. I will leave the decision up to you.

PR #165, opened 2018-12-04
RHIN-1364 Changes for Datepicker to pass refs

yogitattn:
https://rhinogram.atlassian.net/projects/RHIN/issues/RHIN-1364

### What should this PR do?
Changes in datepicker for passing refs so that it's input can be accessed directly anywhere.

NA


alexgsummer:
I left a comment on the ticket which was the cause of this PR...

> Thinking about this some more... If you just need to know if the value has changed or if there is a value set, can you pass an "onChange" type function to the Date component, and then not need a `ref`

yogitattn:
@alexgsummer Refs are required in this case. I have added a comment in rhinofront PR. Please review once

yogitattn:
@juanfabrega I have done the changes as you have suggested, Please review

PR #166, opened 2018-12-05
Juan/rhin 1365

juanfabrega:

### What should this PR do?
Add touch functionality to our tooltip component (before this PR tooltips only worked with mouse hover, which means they didn't work on mobile).

1. Existing functionality for desktop devices should be unaffected. When user hovers over a tooltip trigger, the tooltip will open. When they hover out, it closes.
2. if there is a delay on the tooltip and they hover out before it appears, the timer will be cancelled and the tooltip will never appear. (this was an existing bug that was resolved)
3. On touch devices, the tooltip will open on tap (click event). If a tooltip is open, it will close with a subsequent tap.

Negative.


src/scripts/components/Tooltip.jsx
```diff
@@ -8,6 +8,12 @@ import { TimelineMax } from 'gsap/TweenMax';
 import { UtilitySystem } from '../UtilitySystem';
 
 class Tooltip extends React.Component {
+  constructor(props) {
```
benbruning: Just throwing it out for discussion, in rhinofront we have avoided constructors and gone with es7 property initializer syntax.  I'm not sure if rhinostyle is set to work with those or not, but if so I wonder if we should try to keep it consistent?

src/scripts/components/Tooltip.jsx
```diff
@@ -16,12 +19,22 @@ class Tooltip extends React.Component {
...
+     *
+     * @NOTE: We're not using native touch event sequence since touch devices will emulate a
+     * click event and this is the most reliable way to toggle the tooltip.
+     *  */
     if (!Modernizr.touchevents) {
       tooltipTrigger.addEventListener('mouseenter', this.createTooltip.bind(this));
       tooltipTrigger.addEventListener('mouseleave', this.closeTooltip.bind(this));
     }
+    if (Modernizr.touchevents || Modernizr.pointerevents) {
+      tooltipTrigger.addEventListener('click', this.toggleTooltip.bind(this));
```
alexgsummer: If you define the `toggleTooltip` function using the arrow function, you won't have to bind this function to `this`. 

This line could become:
`tooltipTrigger.addEventListener('click', this.toggleTooltip);`

If you define the function below like:
```
toggleTooltip = (e) => {
  // code here
}
```

The same applies for the other functions where you're doing the same.

src/scripts/components/Tooltip.jsx
```diff
@@ -196,7 +215,15 @@ class Tooltip extends React.Component {
    * @return {void}
    */
   closeTooltip() {
+    // Cancel delayed tooltip open if user hovers out of target. On 'mouseleave'.
+    if (this.tooltipTimer) {
+      clearInterval(this.tooltipTimer);
```
alexgsummer: Technically, you _can_ use `clearInterval` here, but consider using `clearTimeout` for clarity.

src/scripts/components/Tooltip.jsx
```diff
@@ -8,6 +8,9 @@ import { TimelineMax } from 'gsap/TweenMax';
 import { UtilitySystem } from '../UtilitySystem';
 
 class Tooltip extends React.Component {
+  isTooltipOpen = false; // We don't use React state because setState is asyncronous and we need syncronous changes here.
+  tooltipTimer = undefined; // Store setInterval when using tooltip delay. Used in order to clearInterval when needed.
```
alexgsummer: This is the `timeoutId`. Consider updating comment to reflect this as well.

src/scripts/components/Tooltip.jsx
```diff
@@ -206,6 +233,22 @@ class Tooltip extends React.Component {
...
+
+  /**
+   * Open or close tooltip based on its current state.
+   * Used only on touch devices that rely on click events since a click could
+   * call for either opening or closing.
+   * @param {event} e
+   * @return {void}
+   */
+  toggleTooltip(e) {
+    if (!this.isTooltipOpen) {
```
alexgsummer: Personal preference, so leave this as is if you want. Since you're doing an "either or" here and your variable is named `isTooltipOpen`, I find it easier to read like this (notice there is no negation to read):
```
if (this.isTooltipOpen) {
  this.closeTooltip(e);
} else {
  this.createTooltip(e);
}
```

src/scripts/components/Tooltip.jsx
juanfabrega: Good call. I followed the existing code but I like the less verbose approach across the board. 

Will make the changes and merge once complete.

src/scripts/components/Tooltip.jsx
benbruning: @juanfabrega Good point, we could either move to `componentDidMount` or if it needed it before then we'd have to bring back the constructor, which I'm fine with.

juanfabrega: Yup. This was more a comment on future changes to other components. This one doesn't need it.

PR #168, opened 2018-12-07
Juan/rhinostyle docs ie issue

juanfabrega:

### What should this PR do?
Include babel polyfill to fix ES6 object features in IE11. 

1. Rhinostyle docs pages load in IE11.
2. No breaking changes elsewhere.


package.json
```diff
@@ -77,6 +77,7 @@
     "url": "https://github.com/rhinogram/rhinostyle"
   },
   "dependencies": {
+    "@babel/polyfill": "^7.0.0",
```
benbruning: Let's take out the `^`

alexgsummer: I'm not sure if we want to mix babel deps in version 6 (existing babel deps) with the new version 7. Maybe this doesn't matter for polyfill? Just something to verify.

juanfabrega:
@benbruning @alexgsummer changes done for the babel-polyfill import.

PR #170, opened 2018-12-10
Fix UI issues for mobile in charts

rosh-ttn:

### What should this PR do?
Fix UI issues
 ### Jira Link
https://rhinogram.atlassian.net/browse/RHIN-1541
Not Applicable


src/less/charts.less
```diff
@@ -98,22 +98,28 @@
...
-    div {
-      margin-top: @global-whitespace-small !important;
-    }
-
-    .list-title {
-      font-size: @global-font-size-xsmall !important;
+  .chart {
+    .chart__header {
+      display: block !important;
+
```
komalsingh54: no-eol-whitespace

src/less/charts.less
```diff
@@ -98,22 +98,28 @@
...
-      font-size: @global-font-size-xsmall !important;
+  .chart {
+    .chart__header {
+      display: block !important;
+  
+      .chart-doughnut__info {
+        div {
+          margin-top: @global-whitespace-small !important;
+        }
+
```
komalsingh54: no-eol-whitespace

src/less/charts.less
```diff
@@ -98,22 +98,28 @@
...
+      }
     }
-  }
+  }  
 }
 
 //Media queries for big screens
 
 @media (min-width: 1650px) {
   .chart {
```
komalsingh54: no-eol-whitespace

komalsingh54: (block-closing-brace-newline-before)

src/less/charts.less
```diff
@@ -98,13 +98,19 @@
 //Responsive styles for mobiles
 
 @media (max-width: @global-screen-xsmall-max) {
-  .chart-doughnut__info {
-    div {
-      margin-top: @global-whitespace-small !important;
-    }
+  .chart {
+    .chart__header {
+      display: block !important;
```
juanfabrega: Let's make sure !important is always our last resort since they mess up specificity for the codebase moving forward.

In this case the reason it's necessary is that we're using a .u-flex class on the chart__header div in the react component. If you remove that utility class and instead add display:flex in the .chart__header class, you will be able to then change that to a display:block on the media query without needing to use !important.

src/less/charts.less
```diff
@@ -98,13 +98,19 @@
...
 
-    .list-title {
-      font-size: @global-font-size-xsmall !important;
+      .chart-doughnut__info {
+        div {
+          margin-top: @global-whitespace-small !important;
```
juanfabrega: I don't think this !important accomplishes anything. Let me know if I'm wrong here but removing them has no effect.

src/less/charts.less
```diff
@@ -98,13 +98,19 @@
...
+        }
+
+        .list-title {
+          font-size: @global-font-size-xsmall !important;
```
juanfabrega: Same here, no effect removing the !important here.

rosh-ttn: Done

PR #174, opened 2018-12-12
RHIN-1150 dropdown scroll issue

yogitattn:
Issue kink: https://rhinogram.atlassian.net/browse/RHIN-1550

### What should this PR do?
Bug fix for scroll in multiselect dropdown

NA

screenshot from 2018-12-12 16-58-06


src/less/dropdowns.less
```diff
@@ -432,6 +432,7 @@
   z-index: 1;
 }
 
-.multi-select__dropdown__height {
+.multi-select__dropdown__height > div:nth-child(1) {
   max-height: 300px !important;
+  overflow-y: auto !important;
```
mahimaag: @yogitattn Is it required to use `overflow-y: auto !important;` here. As I checkout this branch and dropdown is scrolling properly without adding overflow-y property.

src/less/dropdowns.less
```diff
@@ -432,6 +432,6 @@
   z-index: 1;
 }
 
-.multi-select__dropdown__height {
+.multi-select__dropdown__height > div:nth-child(1) {
```
juanfabrega: So, we've already got this class in rhinofront as well with a different value of 280px. The code here is not an issue, but I don't think added this hard restriction is a good idea in rhinostyle since its only specific to one use case.

Any chance you can chat today about the problem we're having to see if there's a better way to solve it? All other use cases for the dropdown don't have the same issue so I'm confident we can find a way to make this one work too.

yogitattn: @juanfabrega I will just raise a pr in rhinofront and  remove that class from rhinofront side, there is no need of that class there now. And this class will work for every case, and it will remove extra length of dropdown everywhere   wherever we will use this component and scroll will work fine.

juanfabrega: Understood. I don't want to block the progress here and I don't have enough context so I trust your judgement here.

One thing, this new class seems like it only adds a modifier for fixing the height of the dropdown content. Following BEM seems best to change the class name to:

`multi-select__dropdown--fixed-height`

Or something similar to that but using the modifier syntax of `--` that. Let me know what you think.

PR #181, opened 2018-12-21
RHIN-1656 clear dropdown state

yogitattn:
Issue Link: https://rhinogram.atlassian.net/browse/RHIN-1656

### What should this PR do?
Changes to clear state in dropdown when no items selected 

NA


src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -21,6 +21,13 @@ class DropdownMultiSelectAdvanced extends React.Component {
     isViewAllItems: true,
   };
 
+  static getDerivedStateFromProps(nextProps) {
+    if (nextProps.selectedItemsIds.length === 0) {
```
abhicool957:
```suggestion
    if (nextProps && nextProps.selectedItemsIds && nextProps.selectedItemsIds.length === 0) {
```

yogitattn: Actually, This selectedItemsIds prop is a required prop that's why I didn't check this condition here, becoz it will be always there.

PR #182, opened 2018-12-28
RHIN-1643: sticky column not working in iOS devices

mahimaag:

### What should this PR do?
https://rhinogram.atlassian.net/browse/RHIN-1643
This PR provide facility to fix any column either left/right by using an npm module **react-table-hoc-fixed-columns**

First column should be sticky in smart table.


alexgsummer:
The related bug ticket doesn't go into much detail or describe the technical issues. Ideally we wouldn't need another package to solve this, but I'm not familiar enough to offer any direction. I will leave this to @juanfabrega.

adlondon:
> Hey @mahimaag. I found the root cause of this issue and I think we can avoid adding a new dependency and refactoring anything. I've posted more details in the JIRA ticket here, and slack.
> 
> https://rhinogram.atlassian.net/browse/RHIN-1643?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=16561#comment-16561

What's the status on this do we know?

mahimaag:
@juanfabrega as discussed on call, I have tested this sticky column issue on iPhone 5 SE and it is working fine. So you can review this PR now.

PR #186, opened 2019-01-10
RHIN-1714: Optimising unnecessary API calls on Audit logs 

mahimaag:

### What should this PR do?
https://rhinogram.atlassian.net/browse/RHIN-1714

1. Unnecessary *allMembersOrContacts* API calls when open Member/Contact filter dropdown on Audit Logs
2. allMembersOrContacts API should call when typing 3rd text in search field of Member/Contact filter dropdown


src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -49,7 +49,7 @@ class DropdownMultiSelectAdvanced extends React.Component {
   };
 
   handleToggle = () => {
-    if (this.state.isViewAllItems) {
+    if (this.state.isViewAllItems && this.state.searchText.length > 0) {
```
yogitattn: remove this this.state.searchText.length > 0 condition from here and add below

src/scripts/components/DropdownMultiSelectAdvanced.jsx
```diff
@@ -49,7 +49,7 @@ class DropdownMultiSelectAdvanced extends React.Component {
...
       this.setState({ searchText: '' });
       this.props.fetchAllItems('', this.props.filterName.toLowerCase());
```
yogitattn:
```suggestion
if (this.state.searchText.length > 0) {
      this.props.fetchAllItems('', this.props.filterName.toLowerCase());
    }
```

src/scripts/components/DropdownMultiSelectAdvanced.jsx
mahimaag: @yogitattn We need to call fetchAllItems function when searchText is available and it is working fine with these both conditions

src/scripts/components/DropdownMultiSelectAdvanced.jsx
mahimaag: @yogitattn We should clear searchText when searchText is available. Why we need to clear searchText when it is not availble.

src/scripts/components/DropdownMultiSelectAdvanced.jsx
yogitattn: Yes, I know, I just want to move condition to below line, so that, whenever we move from one tab to another, we need to clear the text only, we are doing that already, but this will generate this issue again

src/scripts/components/DropdownMultiSelectAdvanced.jsx
yogitattn: My suggestion is this: 
handleToggle = () => {
    if (this.state.isViewAllItems) {
      this.setState({ searchText: '' });
      if (this.state.searchText.length > 0) {
        this.props.fetchAllItems('', this.props.filterName.toLowerCase());
      }
    }
    this.setState(prevState => ({ isViewAllItems: !prevState.isViewAllItems }));
  };

PR #188, opened 2019-01-24
RHIN-1719 : Fix tooltip issues

rosh-ttn:

### What should this PR do?
Fix issues on tooltips

### Jira Link
https://rhinogram.atlassian.net/browse/RHIN-1719

NA


src/scripts/components/Tooltip.jsx
```diff
@@ -85,8 +87,8 @@ class Tooltip extends React.Component {
 
     $tooltip.setAttribute('id', this.tooltipId);
     $tooltip.setAttribute('role', 'tooltip');
-    $tooltip.classList.add('tooltip');
-    $tooltipContent.classList.add('tooltip-inner', `tooltip-inner--${this.props.type}`);
+    $tooltip.className += 'tooltip';
```
mahimaag: No need to change `$tooltip.classList.add('tooltip');`

src/scripts/components/Tooltip.jsx
```diff
@@ -85,8 +87,8 @@ class Tooltip extends React.Component {
...
+    $tooltipContent.className += `tooltip-inner tooltip-inner--${this.props.type}`;
```
mahimaag: As discussed with juan, he suggested to add classes separately by classList.add like below
```suggestion
     $tooltipContent.classList.add('tooltip-inner');
     $tooltipContent.classList.add(`tooltip-inner--${this.props.type}`);
```

Also add comment before these lines why we are doing this

src/scripts/components/Tooltip.jsx
```diff
@@ -17,6 +17,7 @@ class Tooltip extends React.Component {
    * as well as the context of the currentTarget being incorrect.
    * */
   componentDidMount() {
+    window.addEventListener('scroll', this.closeTooltip, true);
```
juanfabrega: Hey @rosh-ttn  this all looks good, but after reviewing the issue, I now know the root cause, and also know that a scroll listener on the window will not fix it. 

The reason the tooltips move in rhinofront is that rhinofront uses container divs to make it appear that the window is scrolling, when in fact it is only the inner div that is scrolling. This means that the tooltip origin doesn't stay in its place, and it also means that when you scroll in rhinofront the window scroll event WILL NOT fire, so this code won't do anything.

The solution I can think of is that we need to
A. Tie the tooltip more tightly to the parent element.
B. Or pass in the target of the scroll event listener as a prop so that we can listen for scroll on parent div and not just the window.

This is all a bit confusing so if you have any questions, let me know.

rosh-ttn: @juanfabrega For the current solution that I implemented in PR, I tested it  on both mobile and desktop and it works fine. There is no issue in tooltips. Can you please test on your local machine if there is no issue.

Now, for the solution proposed by you, there is an issue that the tooltip do not has any parent div. We are appending the tooltip div directly to the body of the document. Because, there is no parent we will need to append it inside a parent by using the element that triggers the tooltip. But, after I appended tooltip inside the parent element, the tooltip takes up the positioning values, top, right etc. based on the window dimensions, and it do not appears directly on the element which trigerred the tooltip. We need to add the positioning logic too in this case. Also, in cases where the tooltip renders correctly without any such issues, such as the rhinostyle demo, we need to carry on with the existing positioning logic. Attaching screenshot.
screenshot from 2019-01-25 16-34-04

PR #190, opened 2019-02-01
RHIN-1719: Fix tooltip IE issues

rosh-ttn:

### What should this PR do?
Fix fill issue in svg icons which causes tooltip to blink.
Fix tooltip remove function.

Not Aplicable


rosh-ttn:
@komalsingh54 I have raised another PR in rhinofront to solve this issue as it was only on the analytics dashboard.

src/scripts/components/Tooltip.jsx
```diff
@@ -239,8 +239,8 @@ class Tooltip extends React.Component {
    * @return {void}
    */
   removeTooltip(tooltip) {
-    tooltip.remove();
     this.tooltipId = undefined;
+    tooltip.parentNode.removeChild(tooltip);
```
mahimaag: Add comment here why we are replacing tooltip.remove(); to tooltip.parentNode.removeChild(tooltip);

PR #197, opened 2019-02-07
RHIN-1901, 1902 and 1974: UI fixes for react table

rosh-ttn:

### What should this PR do?
Fix UI suggestions.

### Jira Link - 
https://rhinogram.atlassian.net/browse/RHIN-1901
https://rhinogram.atlassian.net/browse/RHIN-1902
https://rhinogram.atlassian.net/browse/RHIN-1974

Not Applicable


src/less/tables.less
```diff
@@ -184,9 +184,9 @@
   -ms-flex-direction: column;
   flex-direction: column;
 
-  .rt-tr.-odd {
+  .rt-tr.-even {
     & + .expand-row {
-      background-color: rgba(0, 0, 0, 0.03);
+      background-color: #f7f7f7;
```
juanfabrega: I know this is a new feature but is there a reason we can't use existing color variables? I see `#f7f7f7` and in the existing variables file we have `@global-color-grey-lightest` which is `#f9f9f9`. Both are almost identical. 

I understand if it was requested exactly that color, but if we have flexibility I'd recommend keeping things consistent whenever possible.

rosh-ttn: @juanfabrega Yes, this can be done. I have updated the color.

PR #198, opened 2019-02-12
RHIN-1974: Remove casing from table data

rosh-ttn:

### What should this PR do?
Remove casing from react table data.

Not Applicable


rosh-ttn:
@juanfabrega By default, there is no casing on existing tables we have in rhinostyle. So, in my opinion, we should keep this consistency for React Tables as well. In the condition where we need specific casing on one/ more columns in table, we can add  specific css rules to achieve the same. Let me know your thoughts on this.

rosh-ttn:
@juanfabrega I have kept the default casing none. To achieve casing as per future requirements, I have used classes to achieve column level casing on one or more of the columns. I have also added a examples inside the react table live code demo. Please have a look.

PR #202, opened 2019-02-22
RHIN-2028: work on dropdown font size and spacing

Sonam-R:

### What should this PR do?
Font size and spacing issue.

1. List

screen shot 2019-02-21 at 5 44 16 pm

screen shot 2019-02-22 at 3 12 30 pm


src/less/dropdowns.less
```diff
@@ -350,6 +350,47 @@
...
+        height: 3rem;
+      }
+    }
+  }
+
+  .form__addon__item {
+    @media (max-width: @global-screen-small-max) {
+      height: 3rem;
+    }
+  }
```
mahimaag: @Sonam-R I think we should create one section for `@media (max-width: @global-screen-small-max)` and add all the classes inside that section. @juanfabrega Your suggestion?

src/less/dropdowns.less
```diff
@@ -350,6 +350,47 @@
...
   padding: @dropdown-padding;
+
+  .resource__wrapper {
+    @media (max-width: @global-screen-small-max) {
+      padding: @dropdown-padding-vertical;
+    }
+
+    .resource__intro__media {
+      @media (max-width: @global-screen-small-max) {
+        width: 3rem;
```
mahimaag: We have a global variable `base-h1-font-size` for 3rem. Use that variable instead of 3rem. Applicable for everywhere.

src/less/dropdowns.less
juanfabrega: Agreed, thanks @mahimaag !

PR #204, opened 2019-02-26
RHIN-2038:  Audit Log - Filter spacing and alignment issues

Sonam-R:

### What should this PR do?
fix dropdown bug in mobile

1. List

screen shot 2019-02-21 at 5 44 16 pm
screen shot 2019-02-22 at 3 12 30 pm


src/less/variables.less
```diff
@@ -80,6 +80,7 @@
 //
 
 
+@global-screen-xxsmall-min: 360px;
```
juanfabrega: @Sonam-R I added this comment in the rhinofront PR but adding here as well just in case. Let's name this breakpoint `nano` instead of `xxsmall` so that we avoid any confusion in the future with the existing `xsmall` breakpoint

Sonam-R: @juanfabrega Done..

PR #207, opened 2019-03-25
Rhin 1392/communication consent

adlondon:

### What should this PR do?
Adds animated "Sliding Radio" buttons for the Communication Consent section for RhinoPay

image
image
image


src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
   }
 }
 
+// add other attr to input to identify it in css file, similar to checked
+
+.rhinoslidingradio__input {
+  &:checked + .rhinoslidingradio__label::after {
+    transform-origin: center center;
+    transform: scale(1.5, 1.5);
+    transition: all 0.25s @global-trans-timing-cubic;
```
juanfabrega: I took some time to review our use of transitions and noticed we've got the following values being used:

0.25s
250ms
0.3s
0.5s
0.1s

This isn't integral to this specific PR but what do you think about creating some variables for `transition-speed-slow` `transition-speed-normal` and `transition-speed-fast`

To try and reign all of these in to some shared standards? Would love to hear your thoughts.

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+      background-color: @global-color-danger;
+      transition: all 0.25s @global-trans-timing-cubic;
+    }
+  }
+  &:checked + .rhinoslidingradio__label.u-text-warning {
+    &::after {
+      border: 2px solid @global-color-warning;
+    }
+    &::before {
+      transform: scale(1, 1);
```
juanfabrega: prefer shorthand `scale(1)`.

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+      background-color: @global-color-warning;
+      transition: all 0.25s @global-trans-timing-cubic;
+    }
+  }
+  &:checked + .rhinoslidingradio__label.u-text-secondary {
+    &::after {
+      border: 2px solid @global-color-secondary;
+    }
+    &::before {
```
juanfabrega: I feel like this code could be condensed a bit if we place the shared attributes outside of the color class bracket. 

Shared attributesfor `&:checked &:before` I see as: `transform` `transition`. That way we're not writing those in three different places, and just changing the `&:after` selector which changes color with each attached color class. 

Let me know if that makes sense.

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+    border-radius: 50%;
+    content: '';
+    display: inline-block;
+    height: 1rem;
+    left: 50%;
+    margin-left: -0.5rem;
+    position: absolute;
+    speak: none;
+    bottom: -.6rem;
+    transform-origin: center center;
```
juanfabrega: prefer shorthand.

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+    transform: scale(1, 1);
```
juanfabrega: prefer shorthand.

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+    width: 1rem;
+  }
+
+  &::before {
+    transform-origin: center center;
```
juanfabrega: prefer shorthand

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+    box-sizing: content-box;
+    height: 3px;
+    width: 100%;
+    transform: scale(0, 0);
```
juanfabrega: prefer shorthand.

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+    content: '';
+    position: absolute;
+    bottom: -2px;
+    left: 0;
+  }
+}
+
+.rhinoslidingradio__container {
+  width: 30rem;
```
juanfabrega: This attribute made me wonder how we make this component re-usable. If I wanted 2 options for example, or 4. How would that work. I know it is not crucial to the current feature but if we ship as is, I suspect it will be hard to find the time to update later. 

Could we pass props for each of the options so that they attached the color classes to each? In that case we could use all existing color variations as well, not just red, yellow, green. 

Would like yo hear your thoughts.

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+}
+
+.rhinoslidingradio {
+  justify-content: space-between;
+  display: flex;
+  border-bottom: 1px solid @global-border-color;
+  & > div {
+    position: relative;
+    min-width: 7rem;
```
juanfabrega: Again thinking about re-usability here. Not sure this min-width is necessary, but want to understand the challenges here. I took it out and there was a slight shift in elements, but nothing major. What I'd like to avoid is using 'locked' words and lengths as an assumption for the options that will go into this component.

src/scripts/components/SlidingRadio.jsx
```diff
@@ -0,0 +1,91 @@
...
+    this.setState({ selectedValue: value });
+    if (this.props.onChange) this.props.onChange(value);
+  }
+
+  getLabelStyle = (value) => {
+    const isChecked = value === this.state.selectedValue;
+    const { options } = this.props;
+
+    let style;
+    if (!isChecked) {
```
juanfabrega: I noted this above, but let's consider making these dynamic based on what is included in the option prop. I'd expect to be able to pass an option that includes both the value as well as the color class I want that option to represent.

src/scripts/docs/FormApp.jsx
```diff
@@ -187,6 +188,17 @@ const FormApp = () => (
           <Radio value="3" label="Radio Three" />
         </RadioGroup>
         <RhinoSwitch label="Switcher" name="exampleSwitch1" />
+        <SlidingRadio
+          selectedValue="2"
+          label="Sliding Radios"
+          name="exampleSlidingRadio1"
+          slidingRadioContainerClass="rhinoslidingradio__container"
+          options={[
+            { label: 'Denied', value: '1' },
```
juanfabrega: Here is where I think we can pass down the color class for each option as part of each option object.

adlondon: 👍

src/scripts/components/SlidingRadio.jsx
adlondon: 👍

src/less/forms.less
adlondon: 👍 Able to set that via 'className' now

src/less/forms.less
adlondon: 👍 removed per our conversation

src/less/forms.less
adlondon: 👍

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
   }
 }
 
+// add other attr to input to identify it in css file, similar to checked
```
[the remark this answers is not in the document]
adlondon: old note to self; removed

src/less/forms.less
```diff
@@ -800,6 +801,97 @@ input[type='checkbox']:focus {
...
+
+.rhinoslidingradio__input {
+  &:checked + .rhinoslidingradio__label::after {
+    transform-origin: center center;
```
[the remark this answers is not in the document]
adlondon: 👍

PR #208, opened 2019-03-26
Juan/input attributes

juanfabrega:

### What should this PR do?
1. Add the ability for our `Input` component to properly pass down custom html attributes such as `data-` and `aria-`. These are currently being ignored since they are not explicitly declared as props in the component.

1. No breaking changes.
2. Any `data-` or `aria-` attributes (passed via the required convention in props) will be rendered to the DOM.


juanfabrega:
> @juanfabrega I don't trust this dangerous code but I'll give you an approval since I owe you one

Your disapproval has been noted @drtonton

PR #211, opened 2019-04-03
Rhin 910/rhinopay icon

adlondon:

### What should this PR do?
Adds the rhinopay icon to the icon list

image


juanfabrega:
Good work! Did you by any chance try a preview of what it looks like at the size that we have it in the prototype? In Rhinofront it's used in a super tiny format, and I'm thinking at this large size it's hard to gauge if the size of the dollar sign works. At first glance I think the dollar sign might need to be bigger for when we shrink it down in rhinofront. See reference here:

https://projects.invisionapp.com/share/7KOMYUK5F3N#/screens/329532489

PR #212, opened 2019-04-05
RHIN-2364:Generalize DateRangeDropdown componenta and move this into rhinostyle from rhinofront

mahimaag:
https://rhinogram.atlassian.net/browse/RHIN-2364

### What should this PR do?
Make **DateRangeDropdown** component generalise by  passing some required props. This was an existing component in rhinofront but we are moving this into rhinostyle now. 

DateRange date picker should work properly for Audit Logs, analytics and closed conversation activity page as working before.


src/scripts/components/DateRangeDropdown.jsx
```diff
@@ -0,0 +1,187 @@
...
+    activeKey: this.props.activeKey,
+  };
+
+  componentDidMount() {
+    this.setState({ isApplyEnabled: false, startDate: moment(this.props.startDate), endDate: moment(this.props.endDate) });
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.endDate !== this.state.endDate && this.props.isCustomDate) {
+      this.endDatePickerInput.datePickerInput.input.setRawValue(this.state.endDate.format('MM/DD/YYYY'));
```
adlondon: This conditional isn't too complicated, so it's not necessary here but consider moving the logic out of the if statement and into a descriptive constant declaration for easier readability

src/scripts/components/DateRangeDropdown.jsx
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  handleDropdownClick = () => {
+    if (this.props.isCustomDate) {
+      const startInputValue = this.startDatePickerInput.datePickerInput.input.element.value;
+      const endInputValue = this.endDatePickerInput.datePickerInput.input.element.value;
+      const { startDate, endDate } = this.props;
+      if (this.state.isApplyEnabled || endInputValue !== moment(endDate).format('MM/DD/YYYY') || startInputValue !== moment(startDate).format('MM/DD/YYYY')) {
+        this.endDatePickerInput.datePickerInput.input.element.value = moment(this.props.endDate).format('MM/DD/YYYY');
```
adlondon: can you make this line a descriptive constant declaration for readability?

`this.state.isApplyEnabled || endInputValue !== moment(endDate).format('MM/DD/YYYY') || startInputValue !== moment(startDate).format('MM/DD/YYYY')`

src/scripts/components/DateRangeDropdown.jsx
mahimaag: @adlondon As per me we should use constant variables for condition check if we are adding same condition multiple times. If we will convert all the conditional logic into constant variable declaration then it will creates many variables. @adlondon @juanfabrega Your thoughts?

src/scripts/components/DateRangeDropdown.jsx
mahimaag: My thought is same as above for this as well.

src/scripts/components/DateRangeDropdown.jsx
```diff
@@ -0,0 +1,189 @@
...
+    maxDate: moment(this.props.maxDate),
+    dropdownMenuItems: this.props.dropdownMenuItems || [],
+    activeKey: this.props.activeKey,
+  };
+
+  componentDidMount() {
+    this.setState({ isApplyEnabled: false, startDate: moment(this.props.startDate), endDate: moment(this.props.endDate) });
+  }
+
+  componentDidUpdate(prevProps, prevState) {
```
juanfabrega: Shouldn't we be checking prevProps here to confirm there was a change in the actual data before triggering a render? I know it's not a hard and fast rule, just wondering what your perspective is here.

mahimaag: @juanfabrega I tested this component. We are getting same values in prevProps and prevState. I don't require this function now as I am managing component re rendering by it's props and state.

src/scripts/docs/examples/DateRangeDropdown.example.txt
```diff
@@ -0,0 +1,57 @@
...
+        isValidDate: true,
+        isDateSelected: true,
+        startDate: date.startDate,
+        endDate: date.endDate,
+      });
+    }
+  };
+
+  render() {
+		const { startDate, endDate, minDate, maxDate, datePickerDropdownLabels} = this.state;
```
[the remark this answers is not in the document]
mahimaag: done

src/less/daterange-dropdown.less
```diff
@@ -0,0 +1,56 @@
+.daterange__dropdown {
```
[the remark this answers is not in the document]
mahimaag: @juanfabrega I tried to manage all css as per your suggestions. Let me know if I miss something. Thanks.

src/less/daterange-dropdown.less
```diff
@@ -0,0 +1,38 @@
+.daterange__dropdown .menu__divider {
```
juanfabrega: Hey team -- sorry, i don't think I was clear here and also did not realize the need for nesting (which falls into the good case I noted in my original comment)

These styles are still nested, even if we have removed the brackets and indentation. The classes will only be applied based on the hierarchy we've noted here. 

While we generally want to avoid that, in this case as I said it makes sense because these are modifying existing classes to fill a customization need. 

So this is my main takeaway here:

1. We should revert back this change, because this is good nesting, and the bracket syntax is easier to follow.
2. This should not be its own stylesheet, it should live inside the dropdowns.less file.

If you see that file you'll notice it has some nested blocks to handle the various types of dropdowns that we have. This is a new type of dropdown so it should fall inside of that existing structure.

Let me know if that makes sense or if y'all have any questions.

Sorry again for the confusion. I hope the intent is clearer this time around.

src/scripts/components/DateRangeDropdown.jsx
adlondon: Still would prefer descriptive constants, but would be fine with a comment in the file to help identify what's going on

src/less/daterange-dropdown.less
mahimaag: done

PR #215, opened 2019-05-14
RHIN-2528: Update variables for old and new field values

rosh-ttn:

### What should this PR do?
Update variables.less file for old and new value colors in the audit logs page

NA

### DIFF
--- a/src/less/variables.less
+++ b/src/less/variables.less
@@ -33,8 +33,8 @@
 @global-color-warning: #f2c746;
 @global-color-warning-light: tint(@global-color-warning, 92%);
 @global-color-emphasis: #fdf9c7;
-@global-color-old-value: #dbedb2;
-@global-color-new-value: #f3aeb2;
+@global-color-green-lightest: #dbedb2;
+@global-color-red-lightest: #f3aeb2;
 
 // Other Colors
 @global-color-black: #080808;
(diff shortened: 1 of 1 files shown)


src/less/variables.less
```diff
@@ -33,6 +33,8 @@
 @global-color-warning: #f2c746;
 @global-color-warning-light: tint(@global-color-warning, 92%);
 @global-color-emphasis: #fdf9c7;
+@global-color-old-value: #dbedb2;
+@global-color-new-value: #f3aeb2;
```
mahimaag: variable name should generic and it should specify the color, not that where we are going to use this variable.

rosh-ttn: Done @mahimaag

src/less/variables.less
```diff
@@ -33,6 +33,8 @@
 @global-color-warning: #f2c746;
 @global-color-warning-light: tint(@global-color-warning, 92%);
 @global-color-emphasis: #fdf9c7;
+@global-color-green-lightest: #dbedb2;
```
juanfabrega: Let's try to follow our current pattern here. 

The green color is a `success` color, so it should go underneath `@global-color-success-light`

Red is `danger` and should go underneath `@global-color-danger-light`.

If we can, let's also use the tint function like in the other cases so that all color variations are not standalone, but instead **derived** from our base theme colors. If this is too difficult to match then we can disregard but I think we should be able to get really close to our target. That way if product decides to modify our theme colors in any way, the corresponding shades will updated as well.

rosh-ttn: @juanfabrega The color family is actually very different from the existing base green and red colors we have, so adding tint is not working out in this case. Hence, I am updating the variable names as shared by you.

juanfabrega: Sounds good, and makes sense!

PR #219, opened 2019-07-05
Hardeep/master/rhin 2893

hardy12994:
*https://rhinogram.atlassian.net/browse/RHIN-2893*

### What should this PR do?
adding smartTable header into rhinostyle

1. List


src/scripts/docs/examples/SmartTableHeader.example.txt
```diff
@@ -0,0 +1,72 @@
...
+  sortTableData = (current, next) => {
+    console.log(current, next);
+    if (current === next) {
+      return 0;
+    }
+    return current < next ? 1 : -1;
+  }
+
+  render() {
+    console.log(this.state.data);
```
mahimaag: remove console log

src/scripts/components/SmartTableHeader.jsx
```diff
@@ -0,0 +1,61 @@
...
+  state = {
+    sortKey: this.props.sortKey,
+    activeCaretClassName: 'u-text-body',
+    headers: this.props.headers,
+  };
+
+  headerHandler = (headerClicked) => {
+    const { headers } = this.state;
+    const oldDirection = this.state.headers[headerClicked].direction;
+    const lastActiveHeader = Object.keys(headers).find(header => (headers[header].default === true));
```
juanfabrega: Is there a better way to manage the properties of the headers so it is clear how to work with them?

1. `default` isn't very descriptive, so it can bee hard to understand what this bool is doing.
2. Direction represented as ints will also create confusion for devs using this component. 

I'm not sure what the best answer is, but it'd be easier to work with this in the future if our direction was a `oneOf` property with values of `ascending` and `descending` the component can handle parsing this data to power the sorting.

Not sure about a better name for `default` let's chat about this one.

src/scripts/components/SmartTableHeader.jsx
```diff
@@ -0,0 +1,61 @@
...
+import React from 'react';
+import {
+  Icon,
+} from '.';
+
+class SmartTableHeader extends React.Component {
+  state = {
+    sortKey: this.props.sortKey,
+    activeCaretClassName: 'u-text-body',
+    headers: this.props.headers,
```
juanfabrega: I'm wondering whether the state should be managed in the parent completely. It is considered an anti-pattern to initialize state from props since it creates multiple sources of truth for the component (props vs state). If all we're doing with this component is setting the class for display based on the sorted column then I think it might be risky to have that live independently of the logic that sorts the actual column data. Let me know if that makes sense.

Basically I'm envisioning the parent just passing down props that tell this component whether the header should display as sorted and in which direction. The onClick function would point to a method passed with props. Not sure if that is better or not. Totally open for discussion. Again my only concern is that this component has an independent state which is handled with different methods than the actual sorting of the data. I guess I also don't see how clicking this thing can sort the data if it is not sending anything to a parent function? I might be missing something, just let me know.

src/scripts/components/SmartTableHeader.jsx
hardy12994: Yeah we can do this as `ascending` instead of default.

hardy12994:
Closing this PR, Creating new one with approach and some modifications.

PR #222, opened 2019-08-06
flipped logic concerning hasChartData

drtonton:

### What should this PR do?
Adjusted logic for a state property in Chart.jsx that was essentially named backwards.

It's `hasChartData`, but its being used as if its `noChartData`. To keep in convention with how we name bools, I just updated the logic surrounding the variable instead of renaming it. 

I also did away with the abstraction of `this.props` into `ChartProperties`. It caused confusion at first glance (true for another dev as well), so I'd just argue that this abstraction causes more issues than having to read `this.props`. No bugs here, just trying to make this a bit more readable.

### DIFF
--- a/src/scripts/components/Chart.jsx
+++ b/src/scripts/components/Chart.jsx
@@ -55,10 +55,7 @@ class Chart extends React.Component {
   );
 
   render() {
-    let hasData = true;
-    if (typeof this.props.data === 'undefined' || this.props.data === null || Object.keys(this.props.data).length === 0 || this.props.data.datasets.length === 0) {
-      hasData = false;
-    }
+    const hasData = this.props.data && Object.keys(this.props.data).length && this.props.data.datasets && this.props.data.datasets.length;
     return (
       <div className="chart">
         <div className="chart__header">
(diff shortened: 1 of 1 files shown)


src/scripts/components/Chart.jsx
```diff
@@ -30,7 +30,7 @@ class Chart extends React.Component {
   componentDidMount() {
```
alexgsummer: I'm thinking we should remove the `hasChartData` from `componentDidMount` since the data might be fetching on initial render. I think you can remove it from state as well and just check if you have data in the `render` function.

src/scripts/components/Chart.jsx
```diff
@@ -66,30 +55,33 @@ class Chart extends React.Component {
   );
 
   render() {
-    const { ...ChartProperties } = this.props;
+    let hasData = true;
+    if (typeof this.props.data === 'undefined' || this.props.data === null || Object.keys(this.props.data).length === 0 || this.props.data.datasets.length === 0) {
+      hasData = false;
+    }
```
adlondon: Consider using const for better readability:

`const hasData = this.props.data && Object.keys(this.props.data).length && this.props.data.datasets && this.props.data.datasets.length`

adlondon: still gross looking but not much you can do there...

drtonton: 👍 done. thanks

PR #223, opened 2019-08-13
RHIN-3080: RhinoPay Manager: UI Adjustments

Sonam-R:
JIRA link:
https://rhinogram.atlassian.net/browse/RHIN-3080

### What should this PR do?
RhinoPay Manager: UI Adjustments

NA

### DIFF
--- a/src/less/dropdowns.less
+++ b/src/less/dropdowns.less
@@ -452,13 +452,9 @@
   text-align: center;
 
   .dropdown {
-    .date-range__buttons {
-      .button {
-        background-color: @global-color-gray-lightest;
-        box-shadow: none;
-        font-size: @global-font-size-small;
-        height: @global-whitespace-large;
-      }
+    .button {
+      background-color: @global-color-gray-lightest;
+      box-shadow: none;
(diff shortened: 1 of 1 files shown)


src/less/dropdowns.less
```diff
@@ -452,13 +452,9 @@
...
 
   .dropdown {
-    .date-range__buttons {
-      .button {
-        background-color: @global-color-gray-lightest;
-        box-shadow: none;
-        font-size: @global-font-size-small;
-        height: @global-whitespace-large;
-      }
+    .button {
```
juanfabrega: Won't this affect nested button classes for all dropdowns on the app?

Sonam-R: I've seen the code.. this style is already given to all dropdown except having class date-range__buttons(i didn't find this class anywhere so missed by someone i guess) and in analytics page, and appointment manager same style is given. That's why I made it globally.

Screen Shot 2019-08-16 at 9 45 27 AM
Screen Shot 2019-08-16 at 9 46 04 AM

PR #226, opened 2019-08-22
RHIN-3180 -Legends are missing on New Inbound Contacts donut chart

vkvisooo:

https://rhinogram.atlassian.net/browse/RHIN-3180

### What should this PR do?
Legends are missing on New Inbound Contacts donut chart

1. List

image-20190821-075918


abhicool957:
@benbruning : This is for rhinostyle  and yaeh branch name is wrong and create confusion

benbruning:
@abhicool957 whoops, my bad.

drtonton:
@vkvisooo I'm pretty unfamiliar with the use of `forceUpdate()`, but based on what I'm reading its generally not recommended unless there are no alternatives, and I can't find any examples in our codebase. Were any alternative solutions considered, and is a re-rendering fixing the issue here?

vkvisooo:
> @vkvisooo I'm pretty unfamiliar with the use of `forceUpdate()`, but based on what I'm reading its generally not recommended unless there are no alternatives, and I can't find any examples in our codebase. Were any alternative solutions considered, and is a re-rendering fixing the issue here?

@drtonton Yes re-rendering fixed the issue cause, this.chartInstance is undefined when the render method is called but after mount this.chartInstance has its value, so we have to call render method again

PR #231, opened 2019-09-06
Mary/develop/rhin 3048

marymhart:
https://rhinogram.atlassian.net/browse/RHIN-3048
Rhinofront PR: https://github.com/rhinogram/rhinofront/pull/3491

### What should this PR do?
Prevents keyboard from covering input when switching tabs on the conversation thread screen on mobile

1. When a user is on a mobile and are looking at the conversation thread screen. When they tap on the RhinoSecure or Note tab, in the compose area, they should be taken to that view and have the cursor auto-focus on the input state.

Screen Shot 2019-09-06 at 12 39 51 PM

BEFORE:

AFTER:
Screen Shot 2019-09-06 at 12 38 28 PM


src/scripts/components/MessageBox.jsx
```diff
@@ -58,6 +58,13 @@ class MessageBox extends React.Component {
     }
   }
 
+  handleFocus = () => {
+    const inboxThreadActive = this.props.messageBoxType ? document.getElementById(`convo__footer__scroll__${this.props.messageBoxType}`) : null;
```
juanfabrega: @marymhart I think this approach is creating too tight a dependency between `rhinostyle` and `rhinofront` since the focus will ONLY work properly based on IDs that are external to this library. Ideally the UI library would be completely agnostic to the app that uses it.

I think a better approach would be to create a `onFocus` prop that can get passed to the `MessageBox` component. That way you can control your scrolling behavior from `rhinofront` and we avoid sprinkling pieces of outside context into our UI library. 

Let me know what you think and if you want to chat about it further.

marymhart: @juanfabrega yes that makes sense! I'll update the PRs

PR #238, opened 2019-09-24
RHIN-3349: Analytics Dashboard | Zeroes are showing in front of graph title when no data

mahimaag:
https://rhinogram.atlassian.net/browse/RHIN-3349

### What should this PR do?
Analytics Dashboard | Zeroes are showing in front of graph title when no data

N/A

image
image


src/scripts/components/Chart.jsx
```diff
@@ -58,7 +58,7 @@ class Chart extends React.Component {
   }
 
   render() {
-    const hasData = this.props.data && Object.keys(this.props.data).length && this.props.data.datasets && this.props.data.datasets.length;
+    const hasData = this.props.data && Object.keys(this.props.data).length > 0 && this.props.data.datasets && this.props.data.datasets.length > 0;
```
yogitattn: @mahimaag I think old line must give same results as new line with new changes becoz: Object.keys(this.props.data).length && Object.keys(this.props.data).length > 0 should work same and should not work for falsy value. Please correct me if i am wrong

mahimaag: @yogitattn If we don't put > 0 it gives me length value 0 i.e an integer not the boolean value and that  in that case `hasData` contains 0 as value. So we are getting 0 infront of title due the below code:
`{hasData && this.props.header && this.props.header.text && (
            <div className={`header__subtitle ${this.props.header.color}`}>
              {this.props.header.text}
              {this.props.subHeader && (
                <span className="subtitle--muted">
                  {this.props.subHeader}
                </span>
              )}
            </div>
          )}` 
 If We put > 0 it will give me boolean value and display JSX according to boolean.

PR #239, opened 2019-09-24
multi-avatars

drtonton:

### What should this PR do?
Creates the `MultiAvatar` component, some styling to go with it, and made adjustments to `ResourceIntro` to handle using the `MultiAvatar` when necessary.

image

1. The component doesn't break
2. The component works
3. The component looks like it does in the prototype


src/less/multi-avatars.less
```diff
@@ -0,0 +1,32 @@
...
+  position: relative;
+  margin-right: 0.8rem;
+}
+
+.multi-avatars__background-avatar {
+  position: absolute;
+  right: 0;
+  top: 0;
+  border: 2.0px solid #fff;
+  border-radius: 50%;
```
juanfabrega: if you are already using the avatar class here, you probably don't need `border-radius: 50%` since that class has that. Same for `overflow` property. It wouldn't hurt to check against the existing classes to make sure we're not adding excess properties to these new classes.

src/less/multi-avatars.less
```diff
@@ -0,0 +1,32 @@
...
+  height: 25px;
+  flex-shrink: 0;
+}
+
+.multi-avatars__foreground-avatar {
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  border: 2.0px solid #fff;
+  border-radius: 50%;
```
juanfabrega: Same as above.

src/less/multi-avatars.less
```diff
@@ -0,0 +1,32 @@
+.multi-avatars {
+  display: inline-flex;
+  flex-direction: row-reverse;
+  width: 40px;
```
juanfabrega: Can we confirm with design/product that these conform to the base size of avatars? In the protoype it seems that the height of the multi-avatar is the same as a regular avatar with the default size. (which i think is 60px).

juanfabrega: Can we confirm with design/product that these conform to the base size of avatars? In the protoype it seems that the height of the multi-avatar is the same as a regular avatar with the default size. (which i think is 60px).

src/less/multi-avatars.less
```diff
@@ -0,0 +1,32 @@
...
+  width: 25px;
+  height: 25px;
+  flex-shrink: 0;
+}
+
+.multi-avatars__foreground-avatar {
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  border: 2.0px solid #fff;
```
juanfabrega: no need for this decimal pixel notation.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,53 @@
...
+class MultiAvatar extends React.Component {
+  renderInitials(initials) {
+    return (
+      <text x="50" y="50" textAnchor="middle" dy="0.35em" fontSize="30">{initials}</text>
+    );
+  }
+
+  renderAvatars() {
+    const { images, names } = this.props;
+    const firstUserIntitials = names[0];
```
juanfabrega: Is this expecting initials or a name? In the original avatar component we pass the full name and then render initials off of that. Would be good to maintain consistency.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,53 @@
...
+    );
+  }
+
+  render() {
+    return this.renderAvatars();
+  }
+}
+
+MultiAvatar.propTypes = {
+  images: PropTypes.array,
```
juanfabrega: Making these an array I think makes prop validation a little problematic and may result in type errors down the line. Basically I can pass any array here and not see an issue, but based on the values in that array I may get all kinds of errors when I try to render the component. Ideally we could pass each prop in separately and type check it, as well as check it prior to attempting to render.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,53 @@
...
+          {firstUserProfileImageUrl === {} && this.renderInitials(firstUserIntitials)}
+        </svg>
+        <svg
+          viewBox="0 0 100 100"
+          xmlns="http://www.w3.org/2000/svg"
+          preserveAspectRatio="none"
+          className="avatar avatar--member multi-avatars__foreground-avatar"
+          style={secondUserProfileImageUrl ? { backgroundImage: `url(${secondUserProfileImageUrl})` } : {}}
+        >
+          {secondUserProfileImageUrl === {} && this.renderInitials(secondUserInitials)}
```
juanfabrega: Not sure under which circumstance the image URL will ever be an empty object, wouldn't it be `null` `undefined` or an empty string if it is not passed?

src/scripts/components/ResourceIntro.jsx
```diff
@@ -135,8 +144,10 @@ class ResourceIntro extends React.Component {
 ResourceIntro.propTypes = {
   avatar: PropTypes.shape({
```
juanfabrega: consider passing `avatar` and `multiAvatar` as their own prop shapes so that we're not mixing avatar props with multiavatar props.

src/less/multi-avatars.less
drtonton: its 40px 
image

drtonton: looks like 40px @juanfabrega 
image

src/less/multi-avatars.less
drtonton: 👍 good catch

src/less/multi-avatars.less
drtonton: updated 👍

src/less/multi-avatars.less
drtonton: updated 🇹🇭

src/scripts/components/MultiAvatar.jsx
drtonton: Soooo, yes and no?

https://github.com/rhinogram/rhinofront/blob/ab695de2b44f9b1c339dd9d575ba1bbe6aa3c461/src/components/ChatInbox.jsx#L48

If you look at ☝️ , and check where we render <Avatar /> in rhinofront, you'll see that in almost all cases we're passing in the initials. Seems that the logic in Avatar.jsx is enforcing that here(basically, accepts either initials or a name):
```
if (name) {
      splitName = name.split(' ');
      initials = splitName[0][0] + splitName[splitName.length - 1][0];
    }
```

I'll add the same check here so it can _expect_ a name, but also accept initials. _That_ would be consistent.

drtonton: done

src/scripts/components/MultiAvatar.jsx
drtonton: @juanfabrega I was thinking of the `style` attribute 🤕 I'll change this.

src/scripts/components/MultiAvatar.jsx
drtonton: 👍 gotcha, this makes sense

src/scripts/components/ResourceIntro.jsx
drtonton: done!

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,61 @@
...
+    );
+  }
+
+  render() {
+    return this.renderAvatars();
+  }
+}
+
+MultiAvatar.propTypes = {
+  foregroundImageUrl: PropTypes.string,
```
juanfabrega: We need to make at least background and foreground **NAMES** required props, here, right? Sorry I missed that the first time around.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,61 @@
...
+    return (
+      <div className="multi-avatars">
+        <svg
+          viewBox="0 0 100 100"
+          xmlns="http://www.w3.org/2000/svg"
+          preserveAspectRatio="none"
+          className="avatar avatar--member multi-avatars__background-avatar"
+          style={backgroundImage ? { backgroundImage: `url(${backgroundImage})` } : {}}
+        >
+          {backgroundImage === '' && this.renderInitials(backgroundInitials)}
```
juanfabrega: I think for this check to work we need to add default prop values to the background and foreground **image** props, since they will be undefined if they are not passed down, so basically with the logic above if you don't pass a image URL, it will also NOT render the initials, so you'll end up with a blank avatar.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,61 @@
...
+          className="avatar avatar--member multi-avatars__foreground-avatar"
+          style={foregroundImage ? { backgroundImage: `url(${foregroundImage})` } : {}}
+        >
+          {foregroundImage === '' && this.renderInitials(foregroundInitials)}
+        </svg>
+      </div>
+    );
+  }
+
+  render() {
```
juanfabrega: Minor thing, you probably can just put everything in the render method. render functions are great when there's lots of JSX blocks that crowd the main render function, in this case we're just calling that one block. This is personal preference so just do what you prefer here.

src/scripts/components/MultiAvatar.jsx
drtonton: yep! good catch. added.

src/scripts/components/MultiAvatar.jsx
drtonton: 🔥

src/scripts/components/MultiAvatar.jsx
adlondon: +1 on this. Personal preference but I try and put most of the JSX in there.

drtonton: Seems fine to me! Updated.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+class MultiAvatar extends React.Component {
+  renderInitials(initials) {
+    return (
+      <text x="50" y="50" textAnchor="middle" dy="0.35em" fontSize="30">{initials}</text>
+    );
+  }
+
+  formatInitials(name) {
+    const splitName = name.split(' ');
```
alexgsummer: You might want to put some protections here if the name isn't in the format you expect.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,62 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+class MultiAvatar extends React.Component {
```
alexgsummer: I don't see the prototype, but is this ever more than two avatars?

drtonton: here's the proto: https://projects.invisionapp.com/share/FRTWGLBBC6T#/screens/380165608

Should just be two, per product

src/scripts/components/MultiAvatar.jsx
drtonton: Unsure what protections to actually place here, now that I think about it. We're doing the same for the `Avatar` component, and it should work with initials and also full names. Open to suggestions and I'll make the change in both components.

src/scripts/components/MultiAvatar.jsx
```diff
@@ -0,0 +1,63 @@
...
+  renderInitials(initials) {
+    return (
+      <text x="50" y="50" textAnchor="middle" dy="0.35em" fontSize="30">{initials}</text>
+    );
+  }
+
+  formatInitials(name) {
+    console.log('dogness', name);
```
marymhart: 👆console.log

drtonton: 🐕

PR #240, opened 2019-09-26
Juan/tag input

juanfabrega:

### What should this PR do?
1. Add a new component for inputting pills in a flexible text area.

1. See --> https://rhinogram.atlassian.net/browse/RHIN-3290


src/less/avatars.less
```diff
@@ -4,7 +4,7 @@
 
 
 @avatar-bg-default: @global-color-gray-light;
-@avatar-bg-member: #7fc0d3; // one-off color that may need to become variable
+@avatar-bg-member: @global-color-teal; // one-off color that may need to become variable
```
juanfabrega:
```suggestion
@avatar-bg-member: @global-color-teal;
```

src/less/avatars.less
```diff
@@ -4,7 +4,7 @@
 
 
 @avatar-bg-default: @global-color-gray-light;
-@avatar-bg-member: #7fc0d3; // one-off color that may need to become variable
+@avatar-bg-member: @global-color-teal;
```
adlondon: 👍teal

PR #241, opened 2019-10-10
Hardeep/master/rhin 3306

hardy12994:

### What should this PR do?
1.Drag n drop feature and reverse too


src/less/variable-messages.less
```diff
@@ -123,3 +169,17 @@
...
+  .category-dropdown {
+    .button__text-wrapper {
+      justify-content: space-between;
+    }
+    .icon {
+      margin: 0 !important;
+    }
+  }
+  .category-variable-list {
+    border: 1px solid #e3e3e3;
```
Sonam-R: use color variable

src/less/variable-messages.less
```diff
@@ -62,12 +63,57 @@
...
+  border-bottom: none;
+}
+
+.category-message__footer__category__active {
+  border-left: 4px solid @global-color-primary;
+}
+
+.category-message__explanation {
+  font-size: @global-font-size-small;
+  font-weight: bold;
```
Sonam-R: Use font-weight:bold(700) variable

src/scripts/components/VariableMessage.jsx
```diff
@@ -18,24 +20,43 @@ import {
 } from '../constants';
 
 class VariableMessage extends React.Component {
+  currentDraggedSpanVariable = '';
+  isCategoryMobileViewVisible = false;
```
juanfabrega: Maintaining an instance property to determine whether to show the mobile view doesn't seem to allow us to react to changes in view width (I.E a user flipping their phone from portrait to landscape will break it). Seems to me this should be done with CSS/LESS breakpoints. Let me know what you think.

src/scripts/components/VariableMessage.jsx
```diff
@@ -18,24 +20,43 @@ import {
...
+    selectedCategory: '',
   };
 
   componentWillMount() {
+    const toUpdateState = {
+      categories: [],
+    };
+
+    if (this.props.isCategoryAvailable) {
+      this.isCategoryMobileViewVisible = window.outerWidth <= 767; //  for mobile view
```
juanfabrega: This relates to my previous comment it will only display the component properly on first load but not react to any changes thereafter.

src/scripts/components/VariableMessage.jsx
```diff
@@ -424,7 +663,9 @@ VariableMessage.propTypes = {
   previewLabel: PropTypes.string,
   name: PropTypes.string.isRequired,
   reset: PropTypes.bool,
-  variables: PropTypes.array.isRequired,
+  data: PropTypes.array,
```
juanfabrega: What's the preference for renaming this from `variables` --> `data`. The new name seems unclear when looking at this component compared to the previous name.

If this prop is required based on the standard function of the component let's mark it that way.

src/scripts/components/VariableMessage.jsx
```diff
@@ -18,24 +20,43 @@ import {
...
   componentWillMount() {
+    const toUpdateState = {
+      categories: [],
+    };
+
+    if (this.props.isCategoryAvailable) {
+      this.isCategoryMobileViewVisible = window.outerWidth <= 767; //  for mobile view
+      toUpdateState.selectedCategory = this.props.defaultSelectedCategory;
+      toUpdateState.variablesOfCategory =
+        this.props.data.filter((item) => {
```
juanfabrega: Seems here we're using a `filter` function as an iterating loop to get properties into an array that is unrelated to the item at hand. Would it not be easer to just filter based on category here and handle adding up the categories in a separate line for clarity?

E.G:
```
const variablesOfSelectedCategory = this.props.data.filter(item => item.category === this.props.selectedCategory);

const categories = [...new Set(this.props.data.map(item => item.category))]; // Set to remove duplicates.
```

src/scripts/components/VariableMessage.jsx
```diff
@@ -18,24 +20,43 @@ import {
...
+        });
+    }
     if (this.props.initialValue) {
-      this.setState({
-        message: this.props.initialValue,
-      });
+      toUpdateState.message = this.props.initialValue;
     }
+    toUpdateState.variables = this.props.data;
+    this.setState(toUpdateState);
```
juanfabrega: In general if all we want to do with `componentWillMount` is to set initial state based on props, then we can do that in the state declaration.

E.G:
```
state = {
  selectedCategory = this.props.selectedCategory || '',
  ...etc
}
```

Moreover if these items are not changing at all, then it would behoove us to pass them as props and just use them, rather than go through the process of initializing derived state. Derived state needs to then be managed everyime props change so it has the potential to become cumbersome for maintainability. At the moment it seems we're ONLY reacting to the `initialValue` prop change but the pattern is not scalable.

src/scripts/components/VariableMessage.jsx
```diff
@@ -316,9 +366,55 @@ class VariableMessage extends React.Component {
...
+    this.currentDraggedSpanVariable = variable;
+    if (event.persist) event.persist();
+    event.dataTransfer.setData('text/html', data.outerHTML);
+    event.dataTransfer.setData('variableId', variable.id);
+  }
+
+  composeMessageDropHandler = (event) => {
+    const variableId = event.dataTransfer.getData('variableId');
+    setTimeout(() => {
+      const text = document.getElementById(`span-${this.currentDraggedSpanVariable.value}`);
```
juanfabrega: Are we able to use a ref here? Unfortunately Ids are not reliable in React when a component is used more than once on a page and can lead to unexpected bugs in the future.

src/scripts/components/VariableMessage.jsx
```diff
@@ -328,37 +424,189 @@ class VariableMessage extends React.Component {
...
 
-  render() {
+  changeCategoryHandler = (category) => {
+    this.setState({
+      variablesOfCategory: this.props.data.filter(item => item.category === category),
+      selectedCategory: category,
+    }, () => this.handleInitValue(this.state.message));
+  }
+
+  categoryMobileView = () => {
```
juanfabrega: prefix render functions with `render` for clarity.

src/scripts/components/VariableMessage.jsx
```diff
@@ -328,37 +424,189 @@ class VariableMessage extends React.Component {
...
+        {showCharacterCounter && (
+          <div title={characterCountTitle} className={characterCounterClasses}>
+            {message.length}
+          </div>
+        )}
+      </div>
+    );
+  }
+
+  categoryWebView = () => {
```
juanfabrega: same as above. Although I'm not sure how this will change if this is handled via CSS media queries. It would be preferable as stated above since it creates a better user experience but also gets rid of a lot of JSX code duplication from these render methods.

src/scripts/components/VariableMessage.jsx
```diff
@@ -328,37 +424,189 @@ class VariableMessage extends React.Component {
...
+        </div>
+        {showCharacterCounter && (
+          <div title={characterCountTitle} className={characterCounterClasses}>
+            {message.length}
+          </div>
+        )}
+      </Fragment>);
+  };
+
+  variableView = () => {
```
juanfabrega: `render` prefix.

src/scripts/components/VariableMessage.jsx
```diff
@@ -18,24 +20,43 @@ import {
...
   state = {
     message: '',
     available: [],
+    variables: [],
+    variablesOfCategory: [],
+    categories: [],
+    selectedCategory: '',
   };
 
   componentWillMount() {
```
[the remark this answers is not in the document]
hardy12994: i think this would be change with getDerivedStateFromProps(), componentDidMount() can't be use because handleInitValue() method need updated state. May be we also call `handleInitValue` it in this.setState callback, instead of getDerivedStateFromProps(). let me know what you think.

src/scripts/components/VariableMessage.jsx
hardy12994: Yeah this looks cool, but acc. to optimization here we are running loop `this.props.data` 2 times. As in above we are doing in single loop in above. Let me know what u think.

src/scripts/components/VariableMessage.jsx
hardy12994: I think dealing with refs here would be difficult. As we don't know list length of variable props are coming from parent component as we have create ref's for all (all variables should be draggable), then we deal it. And also `transformVar()` function creates elements with basic js `document`. 

Maybe Dynamic Refs work. `one ref for all draggable <span's>`, But for that we have to change `transformVar()` that would be big change.

src/scripts/components/VariableMessage.jsx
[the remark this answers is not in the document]
hardy12994: @juanfabrega As this is not the part of draggable thing we can make new ticket to handle this -
Remove old lifecycles `componentWillMount, componentWillReceiveProps` with new lifecycles `getDerivedStateFromProps()`.

src/scripts/components/VariableMessage.jsx
hardy12994: Yeah Sure. Done.

PR #242, opened 2019-10-11
added styling for scrolling in PillsInput input container

drtonton:

### What should this PR do?
Adds `className` prop to the `PillsInput` component so we can alter the styling from Rhinofront as needed.


src/less/forms.less
```diff
@@ -1213,6 +1213,11 @@ input[type='checkbox']:focus {
   padding-bottom: 0.6rem;
 }
 
+.pill-input__container--scroll {
+  max-height: 116px;
```
drtonton: looks arbitrary but was intentional per Benwo

PR #243, opened 2019-10-14
Keaton/pills input format

drtonton:

### What should this PR do?
Allows `PillsInput` to accept a new prop, `inputFormat`, which it passes to the `Input` component.

1. When a filter is selected, it should work :)

A way to test this is to select `Phone` on the filter for a global search with "multi-contact" enabled. Try to enter letters. It shouldn't render any non-numerical input.


src/scripts/components/PillsInput.jsx
```diff
@@ -50,6 +52,7 @@ class PillsInput extends React.Component {
         {renderPills()}
         <div className="pill-input__input-container">
           <Input
+            format={inputProps.format}
```
juanfabrega: With this change it doesn't seem the Input will be able to take in any new props, only format, also we still have old props like `naked`, `name` etc. 

The Input component should basically take `...inputProps` as its only argument and all of these are encompassed in that object.

src/scripts/components/PillsInput.jsx
```diff
@@ -76,11 +79,13 @@ PillsInput.propTypes = {
   initialInputValue: PropTypes.string.isRequired,
   pillType: PropTypes.string,
   pillCloseIconClassName: PropTypes.string,
+  inputProps: PropTypes.object,
```
juanfabrega: This one is probably a requirement no matter what so let's mark it as such.

Also let's add a shape to this object that corresponds to what the inputProps look like rather than an object.

E.G:
```
inputProps: PropTypes.shape({
    name: PropTypes.string.isRequired,
    format: PropTypes.bool,
    onChange: PropTypes.func,
  ...etc
  }),
```

src/scripts/components/PillsInput.jsx
```diff
@@ -76,11 +79,13 @@ PillsInput.propTypes = {
...
 };
 
 PillsInput.defaultProps = {
   pillType: 'powder-blue',
   onPillCloseIconClick: () => {},
+  inputProps: null,
```
juanfabrega: With the input props as a required shape we won't need a default.

src/scripts/components/PillsInput.jsx
```diff
@@ -37,6 +37,18 @@ class PillsInput extends React.Component {
       ))
     );
 
+    const { inputProps } = this.props;
+
+    const propsForInput = {
+      naked: inputProps.naked || true,
+      onChange: inputProps.onChange || this.handleChange,
+      name: inputProps.name || this.props.name,
```
juanfabrega: this is the only one I think we don't need. They're effectively the same so I'd go with `inputProps.name` and get rid of `name` as a prop altogether.

drtonton: 👍

PR #246, opened 2019-10-23
RHIN-3342 changes for Broadcast Message Modal

yogitattn:

Link: https://rhinogram.atlassian.net/browse/RHIN-3342

### What should this PR do?
Minor changes in variable mesage component for broadcast modal

1. List


yogitattn:
> @yogitattn Sorry I'm missing context on this. Can you clarify why we need two separate flags to disable the contenteditable? I don't see the new prop doing anything else, but perhaps I missed it.

@juanfabrega I analysed the code of this Component, Here I can see readOnly(true value) prop will actually hide the variables area but will make the div contenteditable false, but I need the div as conteneteditable true, I am using both the props from Rhinofront to get this functionality

yogitattn:
> I saw your comment and it makes sense, but from hearing that I think the names of these variables are flipped. The `readOnly` prop is hiding the variables area and the `displayMessageAreaOnly` prop is ONLY controlling whether the content is editable or not, but without readOnly it will still show the other section right? IF that is the case, we should flip these so they do what their names say they do.

@juanfabrega I have done the changes regarding  readOnly and displayMessageAreaOnly props, Please review once

yogitattn:
@juanfabrega Please review this Pr

PR #274, opened 2019-12-24
RHIN-969: Worked on BETA label issue

Sonam-R:

### What should this PR do?
Message Translation: Org Preferences - Front End and backend in one

1. List


src/scripts/components/RhinoSwitch.jsx
```diff
@@ -47,10 +47,12 @@ class RhinoSwitch extends React.Component {
...
           />
         </div>
-        <FormLabel className="rhinoswitcher__text" id={this.id}>{label}</FormLabel>
-        {isBetaLabeled && (
-          <span className="title-adjacent-beta-label">BETA</span>
-        )}
+        <FormLabel className="rhinoswitcher__text" id={this.id}>
+          {label}
+          {isBetaLabeled && (
+            <span className="title-adjacent-beta-label">BETA</span>
```
juanfabrega: This  class seems to have gotten deleted from forms.less. do we still need it in the JSX?

Sonam-R: This class is already present in modal.less

PR #275, opened 2020-01-06
RHIN-3370: Removes button box shadow

marymhart:
https://rhinogram.atlassian.net/browse/RHIN-3370

### What should this PR do?

- Removes box shadow from mixin-button-variant & mixin-button-outline-variant
- Updates ResourceIntro to `titleSub: PropTypes.any` to allow string & array

1. Removes box shadow from CTA buttons since this isn't part of the Style Guide.


src/scripts/components/ResourceIntro.jsx
```diff
@@ -187,7 +187,7 @@ ResourceIntro.propTypes = {
   }),
   children: PropTypes.node,
   title: PropTypes.any,
-  titleSub: PropTypes.string,
+  titleSub: PropTypes.any,
```
adlondon: I asked @marymhart to include this in this PR since we use `Arrays` for rendering tags in in the `ResourceIntro.jsx` component when displaying a list of users

PR #279, opened 2020-01-29
added icon

adlondon:

### What should this PR do?
Adds new marketing consent icon and updates the connected parties icon

image

image


src/scripts/docs/IconApp.jsx
```diff
@@ -17,9 +17,9 @@ const IconScope = {
 
 // Add new icons here
 const icons = [
-  'analytics',
   'add-circle',
   'alert-triangle',
+  'analytics',
```
adlondon: Sorted icons alphabetically for easier readablity, as well as added the two new ones

PR #288, opened 2020-04-06
RHIN-4664: Add Data Attributes in the following mentioned page elemen…

kamalpandey484:
PR Link: https://rhinogram.atlassian.net/browse/RHIN-4663

### What should this PR do?
The Data attributes should be added in the following mentioned page elements in order to reduce the timed out failures on elements.

The Data attributes should be added in the following mentioned page elements in order to reduce the timed out failures on elements.
Here we need to add data attribute named (data-cy) for cypress.

### DIFF
--- a/src/scripts/components/DateRangeDropdown.jsx
+++ b/src/scripts/components/DateRangeDropdown.jsx
@@ -97,10 +97,10 @@ class DateRangeDropdown extends React.Component {
   };
 
   render() {
-    const { name, position = 'center', isCustomDate, activeKey, dataCy } = this.props;
+    const { name, position = 'center', isCustomDate, activeKey, dataCypress } = this.props;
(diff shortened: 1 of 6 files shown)


dist/scripts/rhinostyle.min.js
```diff
@@ -29585,7 +29585,8 @@ var Input = function (_React$Component) {
           required = _props.required,
           size = _props.size,
           type = _props.type,
-          validationMessage = _props.validationMessage;
+          validationMessage = _props.validationMessage,
+          dataCy = _props.dataCy;
```
benbruning, line 29589: Like the other pr, does everyone prefer `dataCy` or `dataCypress`?

kamalpandey484: Yeah, dataCypress is more readable, i'm changing it.

